Issue #104, opened 2015-10-25
Justification for closed group


pandurx:
We need to allow users to provide justification/reasoning or additional information when they try to join a closed group. Information will help the group owner to determine who they accept into the group.

Transferred from Leankit T137

Troy-Lawson:
@pandurx was this ever implemented. I thought you were working on this? could be wrong

pandurx:
There is a pull request for this functionality, but it requires upgrade because we revamped GCconnex

pandurx:
the pull request was closed due to new infrastructure change, and not taking in any more enhancement
https://github.com/gctools-outilsgc/gcconnex/pull/107

should we have closed group implemented in new system, we can re-open this ticket

PR #107, opened 2015-10-29
added: functionality to provide justification/reasoning when users tr…

bart-j:
This should be a solution to issue #104. Only thing needed is a French translation, but my French is not that good.... Could someone help me with that? Also not sure if this is the right module to do this, but chose this as this is also the place you overwrite the groups' join action.


pandurx:
Made some small changes to the wording, will be up on pre-production for further testing

ToferC:
Closing issue as last update > 2yrs. Create new issue if still a problem.

PR #110, opened 2015-10-29
fixed: redirect user back after login as this can be the reason for t…

### LINKED ISSUE #100: Redirect does not work
Link from document/email directs user back to the main page after login
Transferred from Leankit cards T199 and T241

bartjkdp:
This hopefully should fix #100. Not sure the problem is solved for all pages but at least the problem is fixed for the pages belonging to the following content types:
- (Calendar) events
- Polls
- Tasks

Also tested the following types, but did not find any redirect issues:
- Activity
- Groups
- Blogs
- Bookmarks
- Files
- The Wire
- Photos

### DIFF
--- a/mod/event_calendar/models/model.php
+++ b/mod/event_calendar/models/model.php
@@ -1884,8 +1884,9 @@ function event_calendar_get_page_content_view($event_guid,$light_box = FALSE) {
 	$event = get_entity($event_guid);
 
 	if (!elgg_instanceof($event, 'object', 'event_calendar')) {
-		$content = elgg_echo('event_calendar:error_nosuchevent');
-		$title = elgg_echo('event_calendar:generic_error_title');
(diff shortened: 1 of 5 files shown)


massierick:
Wow, hope this works out. I've had a lot of people complaining about this, there'll be many happy users once this is implemented!

nschonni:
I looks like the issue was already closed, but this might also be a failure of one of the OWASP top 10 https://www.owasp.org/index.php/Top_10_2010-A10-Unvalidated_Redirects_and_Forwards

bart-j:
I think this is not the case as redirect urls are not defined by a user input, but rather by the Elgg framework. Also I think the redirect URLs are in all cases within the Elgg framework.

Phanoix:
I think this was ported in through other means? Related issue was closed a while ago.

PR #139, opened 2016-09-29
Notification and message summary dropdown

Phanoix:
Merging in the Notification / messages inbox and menu split along with drop-downs with latest unread for each.

Review: @piet0024 @Phanoix


mod/messages/start.php
```diff
@@ -415,6 +415,130 @@ function messages_get_unread($user_guid = 0, $limit = null, $offset = 0, $count
 	return elgg_get_entities_from_metadata($options);
 }
 
+
+/**
+ * Returns the unread messages from other users in a user's inbox
```
piet0024: Do we need this hook in both the missions mod and the wet mod?

Phanoix: Just for us it's not really necessary but if someone wants to replicate this menu split without using our skin then it will be needed.

PR #169, opened 2016-10-19
B extended profile code clean up

Steph4104:


mod/b_extended_profile/views/default/profile/details.php
```diff
@@ -97,18 +93,14 @@
...
-			//error_log("test".array_search($value[0], json_decode($departments, true)));
 			$key = array_search($value[0], $departments);
+
 			if ($key === false){
 				$key = array_search($value[1], $departments);
 			}
-			//error_log('value: '.$key);
-			//echo "lang".get_current_language();
-			//$departments = $meta[0]->value;//array(1, 2, 3);
-			echo elgg_view('input/select', array(
```
Phanoix: You deleted a bit too much here I think.

Troy-Lawson:
This causes a fatal error on profile page.

mod/b_extended_profile/views/default/profile/details.php
Steph4104: Oupss sorry my bad!

PR #282, opened 2016-10-25
Blog required field

Steph4104:


Steph4104:
Also fix PHP warning in line 97

PR #292, opened 2016-10-28
Add content to mesasge

Steph4104:
When you send a message, the email et site message have now content


ToferC:
Just got a few of these. Would it be a caching issue?

Issue #324, opened 2016-10-28
the "Sort Opportunities by" button does not work on the "Archive" page of the Opportunities platform


Troy-Lawson:
GCCON-165
It does work on the "Search Opportunities" page though. 
 
 Archive page: https://gcconnex.gc.ca/missions/main/archive

piet0024:
View pull request tbs-sct/gcconnex#370

PR #370, opened 2016-10-31
Mission archive sort and filter

piet0024:
#324 

Users should now be able to sort and filter archived missions.


piet0024:
I have tested this on preprod and @ethanWallace has confirmed it worked on preprod.

Issue #417, opened 2016-11-03
Suggested colleagues bugs out with unvalidated users


Phanoix:
Suggested colleagues query fetches unvalidated users as well as validated ones, which occasionally cause pages that contain the module to error out or come back blank.

Phanoix:
I will add a quick check to ensure the suggested users are validated, refactoring of the feature to follow.

PR #422, opened 2016-11-03
Suggested colleagues bug quick fix

Phanoix:
Will no longer crash when trying to show unvalidated users - it will now simply skip them. The feature is set to be rewritten and will handle these cases more elegantly at a later time.


Phanoix:
Fixes #417 . This patch is already in production.

Issue #413, opened 2016-11-02
Better validation/feedback to user when submitting information on user profile


ethanWallace:
Provide better feedback to the user when they submit information into their profile, mainly in education/work experience. Goal is to highlight the fields that have errors not submit and delete already entered information the user has entered.

ethanWallace:
Pull Request #423

PR #423, opened 2016-11-04
Better validation on profile education/work experience

ethanWallace:
Form will no longer submit if the user has entered incomplete information(missing field) or incorrect information (end year is less than start year). Offending fields will be highlighted in red and a message will appear above save button to address the errors.


ethanWallace:
LInked to Issue #413

PR #469, opened 2016-11-15
Mission Cards should now display the proper language for applicants

piet0024:
If there is only one applicant the language should now say applicant instead of applicants.

Hey @Steph4104 can you check my code really quick and let me know if you feel this merge request is good to? :)


Steph4104:
I don't remember if somewhere it's say that the male is use for female... Like, in french we have Candidat for male and Candidate for female (Candidat = applicant) So if we only have one applicant female, we should see candidat**e** not candidat... So I'm not sure if we want to put the (e) at the end. Otherwise, everything look perfect!

Issue #456, opened 2016-11-14
Change button for tabs - Content translation


Steph4104:
Change all button for translation content to tabs

PR #473, opened 2016-11-15
Change button tabs

Steph4104:
#456


Phanoix:
@LemieuxGen Since we haven't communicated much about the whole feature being released I guess this will just be part of the initial launch for the feature.

Issue #461, opened 2016-11-14
Set Create Opportunity default to user department from profile


ToferC:
When creating an opportunity in the Opportunity Platform, the default department should be set to the department from the user profile.

PR #478, opened 2016-11-16
Missions_organization fix for pre-filling department field in opportunity creation

Phanoix:
addresses issue #461  
This functionality was disabled before due to it not working properly and defaulting the department drop-downs to less than useful values, which combined with and IE bug prevented users from easily filling out the form. With this fix it should hopefully work as intended.

@piet0024 could you quickly look over this? It ended up only being a few lines
@LemieuxGen This is probably notable enough to require some level of communication


LemieuxGen:
@Phanoix Thanks for the heads up!

PR #487, opened 2016-11-21
now receiving notifications about likes

pandurx:


piet0024:
👍 I got a notification for a like on preprod

PR #488, opened 2016-11-21
Sticky topic

pandurx:


piet0024:
Editing forum topic -> Language for create sticky topic not set on the form

PR #490, opened 2016-11-22
fixed issues with auto-populate when sending msg to non-colleague

pandurx:


Phanoix:
looks like this change is replicated in #488

PR #501, opened 2016-11-22
update git ignore and add saml link mod for use in simple sign on

Troy-Lawson:
Initial saml_link repo update


Troy-Lawson:
duplicate

PR #504, opened 2016-11-23
Add flag content option in dropdown of contactform

Steph4104:
#421


piet0024:
👍 - Like the yellow info box giving more info when you choose the flag content option. 

Perhaps the language would make more sense if its said something like "Report content or behaviour" as 'flagging' might cause confusion.

Steph4104:
Miss link in the content side under FAQ. It's in the branch but not in preprod.

PR #531, opened 2016-11-28
Fixed other characters displaying instead of french characters

### LINKED ISSUE #524: French accents displayed as other caracters in User's name when viewing messages/notifications
When a message is opened (from the inbox or sent messages) or a notification, the French accent a user's name and/or last name is replaced with other characters (see attached screen shot). Note that this is ONLY occurring in the page where the message is displayed, not in the page where the list of messages or notifications are displayed (ex: inbox); it's fine in the list pages

gcconnexaccentissue
.

ethanWallace:
Fixes #524


ethanWallace:
Fixes issue #524 French characters should now appear properly when reading a message.

PR #535, opened 2016-11-28
Gcconnex theme - Moving GCconnex related branding and language into it's own mod

piet0024:
Hey @Steph4104 can you pull this branch and double check this doesn't break anything? 👌


lucbelliveau:
After a fresh install, I get the following error in the plugins configuration page:

> This plugin is invalid: This plugin's directory must be renamed to "gc_gcconnex_themee" to match     the ID in its manifest.

Issue #308, opened 2016-10-28
Group Cover Image - Broken


Troy-Lawson:
GCCON-255
Since the last update on GCconnex (October 23), the group cover image is now broken. The images are no longer showing up (see attached screen shot).

LemieuxGen:
This bug still persists in the system. Any updates? or is this part of a sprint?
![broken group cover image](https://cloud.githubusercontent.com/assets/15378977/20568853/d98ccda8-b16b-11e6-9c04-06d6519ec6d2.JPG)

PR #554, opened 2016-12-01
Cover photo transferring bug

ethanWallace:


ethanWallace:
Fixes issue #308 Cover photos would break after transferring group ownership

PR #555, opened 2016-12-01
Docker

lucbelliveau:
A good start on using Docker.  Follow the instructions in [README.md](https://github.com/tbs-sct/gcconnex/blob/db479396bde2690e2793a2d945401ada717c185d/README.md) to get going.

A quick note about volumes...  The "data" folder elgg uses is mounted to ./data - when you bring up the containers if that folder does not exist, it will be created and owned by `root`.  To prevent that, I added the folder to git to make sure it is owned by your current user.  (I suspect this isn't a problem on Windows)

Files created inside that folder by apache inside docker however will be owned by apache's UID inside docker... if you cannot delete them as a regular user, a good trick is to use docker itself to run commands, for example to wipe the data directory you could run:

    docker-compose run --rm gcconnex rm -rf /data/*

This command would delete all files in `/data` from the point of view of the `gcconnex` container, which is mapped to your sourceroot/data folder by `docker-compose.yml`.  Of course alternatively you could simply execute `sudo rm -rf ./data/*`

I'm happy to provide more info if neeed.


ToferC:
This is fantastic @luc-belliveau. Looking forward to trying it out. Thank you!

PR #577, opened 2016-12-06
Updated blank labels to "Other"

### LINKED ISSUE #500: Opportunities Platform - Error in visualizing stacked graph by program area
One field of the stacked graph is lacking a tag - bottom of the graph - dark blue.  Probably "Other".

lucbelliveau:
Stack graph will now use the label "Other" instead of an empty string when separated by:
1. Program Area
1. Opportunity Type
1. Reason to decline
1. Location

fixes #500


ToferC:
Thanks, @lucbelliveau - We are getting an error on pre-prod with this, but it looks like it should work on production.  Will advise after code push this weekend.

lucbelliveau:
Ok... what is the error?  Just in case...

Phanoix:
The developer tools mod was adding html comments to the JSON string causing the ajax call to error out, it's working on pre-prod now.

PR #581, opened 2016-12-07
Asynchronous invitation of found candidates enabled.

### LINKED ISSUE #528: Opportunities Platform - Ability to share to more than one user from Find/Member Search results
As an opportunity creator, I want to be able to go down the list of members and offer opportunities to more than one user without having to re-run the search.  Options for resolution: check-box beside members and a single share opportunity button, ability to click "Share Opportunity" button and remain on search results page.

lucbelliveau:
Also fixed reversed status message.  ie You have shared the opportunity 'username' with 'opportunity'.

fixes #528

### DIFF
--- a/mod/missions/actions/missions/invite-user.php
+++ b/mod/missions/actions/missions/invite-user.php
@@ -65,7 +65,7 @@ if($err == '') {
 	
 	//add_entity_relationship($mission->guid, 'mission_tentative', $applicant->guid);
 	
-	system_message(elgg_echo('missions:invited_user_to_mission', array($applicant->name, $mission->job_title)));
+	system_message(elgg_echo('missions:invited_user_to_mission', array($mission->job_title, $applicant->name)));
 	
 	forward(elgg_get_site_url() . 'missions/main');
 }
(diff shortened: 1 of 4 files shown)


ToferC:
Hi @lucbelliveau - @Phanoix has pushed your commits to pre-prod, but the multi-select isn't working on our system.  Can you touch base with @Phanoix re: possible reasons for this?  Thanks!

lucbelliveau:
Sure... is there an error @Phanoix ?

Also were the Elgg cache flushed?  Because of [ajax.js](https://github.com/tbs-sct/gcconnex/blob/967bf32fe2c71392681c53c4c9c9da867670a75c/mod/missions/views/default/js/mission-invite/ajax.js) the caches must be flushed...

ToferC:
@Phanoix - can you confirm this?  It isn't working for me on pre-prod.  From Share Opportunity page, it redirects to Search for Opportunities after I click the "Share" button.  Thanks!

Phanoix:
I just tested it using the regular members search and it worked (on pre-prod).

ToferC:
This is live and works great from the Find Candidates Screen. @Phanoix - can you please enable the same functionality from hitting the Find button on an opportunity?  Thanks!

Phanoix:
Looks like the that list works differently, using links instead of forms with buttons for the invitation action so the ajax invitation script will need to be adapted for that.

ToferC:
Gotcha. @lucbelliveau - can you take point on that?

lucbelliveau:
Sure - should I reopen this PR or create a new one?

ToferC:
Let's say new one to keep it clean.

Issue #532, opened 2016-11-28
Breadcrumbs - Discussion leads to 500 error


ToferC:
Selecting the Discussion item from the breadcrumbs Home -> Discussion -> <GROUP> -> Discussion Topics (GCconnex.gc.ca/discussions/all) returns a 500 error.

Expected link would be to group discussions.

Perhaps re-organize breadcrumbs to be: Home -> <GROUP> -> Discussion -> <Topic>

PR #594, opened 2016-12-12
Discussion page now loads

ethanWallace:
Sidebar was checking for a page owner and not finding one. Caused page
to error.


ethanWallace:
Fixes #532

PR #609, opened 2016-12-13
Pan/zoom functionality in Opportunities Platform's stacked graph analytics.

### LINKED ISSUE #499: Opportunities Platform Analytics - Improve Visualization Rendering
Users should be able to scale and view histogram and bar graph analytics visualizations on a single screen.  Graphs should scale to allow data to render on screen and should be viewable on mobile devices.

lucbelliveau:
- Change data from service to return time series instead of string labels,
- Added pan/zoom via mouse drag and doubleclick; as well as control buttons visible on top of the graph,
- Updated size of graph to match window size (on resize as well),
- Added no whitespace CSS rule to data table headers to improve legibility,
- Tested on Mobile, full graph is visible and pan/zoom button controls work well.

Requires cache clean due to new language strings.

fixes #499


ToferC:
This looks excellent, Luc.  Thanks!

Issue #537, opened 2016-11-28
Hyperlink tags to search


ToferC:
When clicking on any site "tags", the user is brought to the old search page which yields no results.  Clicking a tag should bring the user to a GSA search of that tag, or (even better) to a list of content that also has that tag.

pandurx:
https://www.google.com/support/enterprise/static/gsa/docs/admin/72/gsa_doc_set/xml_reference/request_format.html#1089494

GSA should be able to pick up the meta name="keywords"

KevinChagnon:
When clicking on any site "tags", the user is brought to the old search page which yields no results.

PR #630, opened 2016-12-20
Federated Search update

pandurx:
The tags are now implemented as Metadata, GSA should now be able to pick that up and users should be able to do seach by tags (via tags in content) using the GSA. 
Those unusual hyperlinks have been corrected (display only) but content is not updated, this usually affects older data)


pandurx:
this branch deals with #537 and issue 13 (in gccollab)

PR #632, opened 2016-12-21
Streaming wire page

piet0024:
If a user stays idle on the wire page it will notify them that there are new posts. It will then load the new posts through Ajax


ToferC:
Just tested on pre-prod.  Still grabbing an extra wire post and duplicating it.  Also. please ensure that the streaming wire is also applied to the Newsfeed page.  Thanks!

Phanoix:
The newsfeed is a very different feed and is most users' home page, so while some of the work done here can be reused we should approach a streaming newsfeed as a separate piece of work. It can't use the current API to fetch the feed and we'll need to consider server load impact more.

ToferC:
Agreed on newsfeed. To clarify: streaming wire should also be on newsfeed
page.

piet0024:
I made some changes to this so it should no longer create duplicates. Before it was comparing post text and now it is comparing GUIDs. The mod now also has a custom index widget for the newsfeed page that should function the same as the wire page.

ToferC:
Tested and confirmed it works. On occasion the whole page will refresh
instead of just the widget. Also, the pop-up of new wire items is a bit
sudden. Might want to look at a smooth transition if it's not too much
work. Thanks!

PR #679, opened 2017-01-12
En/fr content rework

Phanoix:
This does not change the user experience, only how the different language versions of content are stored in the database.


Phanoix:
Rebased, resolved conflicts.

Phanoix:
We should make sure the max length check either counts the en/fr parts separately or double the cutoff when both languages are present (not ideal, people might find out and game the system by adding a single character to the other language to get double the allowed length)

mod/tidypics/actions/photos/album/save.php
```diff
@@ -49,10 +49,10 @@
 $album->access_id = $access_id;
 $album->title = $title;
 $album->title2 = $title2;
-$album->title3 = gc_implode_translation($title,$title2);
+$album->title = gc_implode_translation($title,$title2);
```
nschonni: should this move up instead of setting this twice in the same block? Description is doing the same below

mod/wet4/views/default/object/elements/summary.php
```diff
@@ -65,7 +62,14 @@
 	$tags = elgg_view('output/tags', array('tags' => $entity->tags));
 }
 
+
+if ( !$title_link && json_decode($entity->title) ){
```
ethanWallace: This line adds an additional header of the content title to the page. It appears right above the author details.

nschonni:
It might be easier to do separate PRs per-Mod instead of a long running PR that's more likely to hit conflicts

klecuyer:
@Phanoix here?

Submitted on behalf of user: en/fr labels appearing in Group titles
image

klecuyer:
image
Also displaying in discussion content preview, but not in actual discussion or discussion title

PR #713, opened 2017-01-19
issue#685 gsa fedsearch results need bilingual titles

pandurx:


pandurx:
Due to the recent issues with the search, this will be implemented once the search results are OK

PR #716, opened 2017-01-19
allow gsa to crawl files

pandurx:


pandurx:
I have tested and the metatag for noindex does not appear in the source code for the File pages

PR #773, opened 2017-02-03
Opportunities add data

### LINKED ISSUE #498: Opportunities Platform Analytics - Additional Data Required
Currently tracking Posted, Cancelled, Completed.

Should also track:

- Offered (when offers made)
- Declined  (when offers declined)
- In Process (when offers accepted)

lucbelliveau:
fixes #498


ToferC:
I'll review first thing Monday. Thank you!

ToferC:
@Phanoix - can we bring a copy of the missions data into pre-prod?  I'm seeing something, but not sure if the data graph is complete because of the spotty pre-prod data.  Thanks!

lucbelliveau:
Keep in mind that only "declined" will be accurate for past data; as discussed in #498 the other ones needed new data to be tracked.. so only events that occurred after this PR would be visible.  To test you would need to create and action new opportunities.

Phanoix:
Generating a chart spanning about a year and grouping by month is probably the best we can do with what we have there now, another thing we can do is check it on dev which should have more recent data. We can also look into updating pre-prod at some point, but it would probably be easier to check on dev right now.

ToferC:
Thanks to both of you.  @Phanoix - no need to update the DB right now.

I've created a new opportunity and had @LemieuxGen  apply to it.  Summary of testing:

- Opportunity posted = all good
- Opportunity offered = displaying after offer made
- In Progress = not displaying after applicant accepts opportunity

I then removed the applicant from the opportunity and stacked graph displays 3 items: Posted / Offered and In Progress.

Applicant re-applies to process, I offer the opportunity again.  The stacked graph displays 4 items: Posted / Offered x2 and In Progress.

The applicant declines the opportunity.  The stacked graph displays 5 items: Posted / Declined / Offered x2 and In Progress.

The applicant re-applies to the opportunity and I offer it to them again.  The stacked graph displays 6 items: Posted / Declined / Offered x3 and In Progress.

I complete the opportunity.  The stacked graph displays 7 items: In Progress x2 / Declined / Offered x3 and Completed.

I think that in general this looks good as we're tracking transactions, but the "In Process" isn't displaying when the opportunity is offered and accepted, but after another transaction occurs.

We should probably also change the way that Posted/Completed/Cancelled are counted so that we're using the same standard of transactions (right now it seems that they are separate switches, i.e., that an opportunity can be one of the three at any point in time).

Suggest we also change the stacked graph "Separate Opportunities by" to: "Transactions" in FR/EN.  Also, if possible, we should create an info bullet above the graph stating: "Number of transactions on opportunities" in EN and "Nombre de transactions sur les possibilités" en français.

Titles would actually be a good idea for all graph types.

lucbelliveau:
I fixed the issue with the "In Progress" when it's in the current period; and I renamed the graph to "Transactions".

@ToferC could you clarify what you meant by this:

> We should probably also change the way that Posted/Completed/Cancelled are counted so that we're using the same standard of transactions (right now it seems that they are separate switches, i.e., that an opportunity can be one of the three at any point in time).

Thanks

ToferC:
The new data elements are count (), while it appears the old ones Posted,
Completed, Cancelled are states.

On the graph we should ensure we are using a common and understandable
measure.

So - for this graph we should count the number of  Posted, Completed,
Cancelled opportunities.

Make sense?

lucbelliveau:
I think so... you're saying if an opportunity moves from posted to completed, the graph should show 2; one in each state?  And right now it would only show the final state of the opportunity... is that right?

ToferC:
That's it.

ToferC:
@lucbelliveau - please let us know when you've updated this pull request or Issue #498.

ToferC:
@lucbelliveau - just a quick follow up on this pull request.

lucbelliveau:
Sorry for the immense delay, but this is finally ready to be tested again.

Phanoix:
Thanks, I'll put it up on pre-prod next time it's updated.

ToferC:
Just reviewed on preprod.  We're getting a 200 error: preprod.gcconnex says: 200
SyntaxError: Unexpected token < in JSON at position 1

lucbelliveau:
Can you check if there is a PHP error being generated by the ajax call?

Phanoix:
just updated preprod

Phanoix:
it's the developer tools mod inserting comments into the json and causing errors

ToferC:
Tested again. Seems to be working. Will test all elements tomorrow AM.
Thanks!

Phanoix:
Does this look good to go for the update this weekend?

ToferC:
Phil put an update in the associated issue. One item still not working.

lucbelliveau:
I don't see any updates in #498 ?

ToferC:
He put it in #1092. Sorry!

ToferC:
@Phanoix - let's put this in the update and flush the cache. At minimum we'll start collecting the right data. Thanks Ilia and @lucbelliveau

PR #781, opened 2017-02-10
change machnine translation to accept jsonp

Troy-Lawson:
change in machine translation to accept new format and jsonp response format


Troy-Lawson:
this is ready to go, and has already been implemented in production.

PR #806, opened 2017-02-22
strips non relevant content if agent string is gsa-crawler (string co…

pandurx:
…mpare, substring)


pandurx:
@LemieuxGen tagging you here for updates

Issue #817, opened 2017-02-27
wrong text for button "Add new event" in the Event calendar page


LemieuxGen:
In the Evenet Calendar page: https://gcconnex.gc.ca/event_calendar/list/ 
The button to add a new event was change to show encoding, in both English and French.
Should say "Add new event" (FR: Ajouter un événement)
![calendar_button_add new event_fr](https://cloud.githubusercontent.com/assets/15378977/23362851/ae59a0ac-fcc5-11e6-8e34-52daa4a4719f.JPG)
![calendar_button_add new event](https://cloud.githubusercontent.com/assets/15378977/23362852/ae62b3ae-fcc5-11e6-82ce-301897077fde.JPG)

PR #823, opened 2017-02-28
Correct text in  add event.

Steph4104:
Remove 1 from the string

#817


mod/wet4/views/default/event_calendar/show_events.php
```diff
@@ -72,11 +72,10 @@
 	}else{
 		$new_link = elgg_view('output/url', array(
         'href' => "event_calendar/add",
-        'text' => elgg_echo('event_calendar:new1'),
```
piet0024: Do we know why the 1 was there? Was there supposed to be a new string?

Steph4104: No, I look in all the site and event_calendar:new1 do not exist. 
Also, we never talk about changing the wording "add event" for something else.

PR #828, opened 2017-03-02
More gsa issues

pandurx:


pandurx:
before: http://gcconnex.gc.ca/blogs/view/1234/some-blog-title
after: http://gcconnex.gc.ca/blogs/view/1234/

PR #832, opened 2017-03-06
GCTools Recommender supporting API implementation

lucbelliveau:
This is for the GCTools recommendation project - We need an API on production that will allow us to export required data.  This PR implements both "user" and "opportunity" data exports.

The API is secure and can only be accessed by a client who knows the secure private key located at NRC.

This update does not impact any existing functionality.  

@cdegrass @ToferC  Can we expedite its release to production?

### DIFF
--- a/mod/missions/api/v0/api.php
+++ b/mod/missions/api/v0/api.php
@@ -1,0 +1,279 @@
+<?php
+
+/**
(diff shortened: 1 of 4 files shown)


ToferC:
@troylawson - can you review and merge? Thanks!

PR #837, opened 2017-03-08
removes description unless it contains no html

pandurx:
This has been fixed in Notification Digest. We will push this if Notification Digest is decided to be delayed for production implementation.


pandurx:
any notification that contains content that has <html><table> tags are removed entirely

Issue #807, opened 2017-02-22
Improving accessibility for the visually impaired


AlisonRumsey:
The following are a few issues that could improve accessibility for the visually impaired: 
1. JAWS (screen reader) is unable to select the text box when adding a file. 
2. JAWS does not pick up on the Welcome tutorial that pops up upon initial sign in. 
3. JAWS does not always pick up on the "edit" field. ie you must click instead of using regular JAWS functions (similar to using tab)
4. Ctrl+v works in “Files” widget, but right click paste does not. Both work in “Blogs” widget. 
5. When using JAWS, and clicking on the "Messages" or "Blogs" tab, the page does not load.

#BreakingBarriers

ToferC:
@Troy-Lawson - would be great to get these items into the next code sprint.

PR #852, opened 2017-03-16
Changed Onboarding welcome modal to get picked up by screen readers

piet0024:
Related to issue #807


mod/gc_onboard/views/default/page/elements/step_counter.php
```diff
@@ -13,6 +13,7 @@
 $total_steps = elgg_extract('total_steps', $vars);
 $current_step = elgg_extract('current_step', $vars);
 
+$screen_reader = elgg_format_element('div', array('class'=>'wb-invisible'), 'Step '.$current_step. ' of '.$total_steps);
```
Phanoix: Shouldn't the "Step x of y" be a language variable and get translated?

Issue #798, opened 2017-02-20
Title/Description of posting in French section of Notification showing in English


KevinChagnon:
The messages for new discussion posts only seem to show the English side of the post, not the French, even though French topic message was added to the French side and the English topic message was added to the English side when creating the post. Another issue is that only the English title shows up in both the discussion forum and the email messages even if a French title was added to the post via the French title field. 

![messagesmissingfrench 2](https://cloud.githubusercontent.com/assets/23527358/23138747/1d15a3dc-f777-11e6-9678-5ec81ee0c089.png)

PR #863, opened 2017-03-22
French in new discussion notification

Steph4104:
Notification for new discussion show french title and french discussion


Steph4104:
Pull request related to issue #798

Steph4104:
Add the change, now, if the variable of the language is empty, it put the other one content

Issue #841, opened 2017-03-10
Email Members: Number of recipients not matching number of group members - Error when sending msg


KevinChagnon:
Group: https://gcconnex.gc.ca/groups/profile/16909638/leadership-for-all

When the option to Email Members is selected, the system automatically applies a checkmark beside all the member's names within the list.  The "number of recipients" shows the same number as the number of group members;

![group members number](https://cloud.githubusercontent.com/assets/23527358/23803916/1b8d4048-0586-11e7-88e6-80503691bc84.JPG)

![email members checked](https://cloud.githubusercontent.com/assets/23527358/23803847/d5331a32-0585-11e7-8d8b-4653dc2962b5.JPG)

However, if the checkmarks are cleared using the button at the bottom of the page (Clear selection) and then "All Members" is selected, the number of recipients changes to a higher number that does not match the number of group members;

![email members - all members](https://cloud.githubusercontent.com/assets/23527358/23804011/73b17a0a-0586-11e7-8259-bf5e219f6020.JPG)

When the group owner attempts to email All Members, she gets an error stating "Invalid input to perform this action" ;

![email members error](https://cloud.githubusercontent.com/assets/23527358/23804075/ac7dc6f4-0586-11e7-8b12-66a9cb3ca369.png)

Screenshot of error message was sent to me when there were fewer group members.

User: "If I select one person, it works, but when selecting “All Members” it does not go through (error)"

KevinChagnon:
Additional info:

**Q**. You mentioned that you can send individual messages, have you tried sending a message to multiple (but not all) members? 
**A**. Yes. It seemed to go through when I tried one page of 250 members, but not “All members”

**Me**: You also noticed that the number of recipients did not match the number of group members.  I came to determine that this oddity only happens if you clear the selection of “all members” – which is the default when you click “Email Members” - and then reselect “All Members” 
**User**: I noticed that the number is higher when I tick the “colleagues” box, which I assume includes my own colleagues. When I uncheck that box, the number seems accurate.

**Q**. Did you get your error message when trying to email all group members right after you had clicked “Email Members” (the default selection) or after you had un-selected and then re-selected All Members ? 
**A**. I tried both and neither worked.

**Q**. Does this happen when you try to Email Members within other groups you own ? 
**A**. I’ve not tried other groups that I own, although since the 250 members worked, my other group is under 100.

[...]

PR #869, opened 2017-03-24
Number of recipient fix

Steph4104:
Should fix the number of recipient when you are using the all member button or the My colleague checkbox in the next push.

Related to #841

### DIFF
--- a/mod/group_tools/views/default/js/group_tools/mail.js
+++ b/mod/group_tools/views/default/js/group_tools/mail.js
@@ -29,11 +29,19 @@ elgg.group_tools.mail_clear_members = function() {
 		$(this).removeAttr('checked');
 	});
 
+		$('#group_tools_mail_member_selection input[name="select_all"]:checked').each(function() {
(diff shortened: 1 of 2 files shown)


mod/group_tools/views/default/js/group_tools/mail.js
```diff
@@ -28,6 +28,14 @@ elgg.group_tools.mail_clear_members = function() {
 	$('#group_tools_mail_member_selection input[name="user_guids[]"]:checked').each(function() {
 		$(this).removeAttr('checked');
 	});
+    
+    	$('#group_tools_mail_member_selection input[name="select_all"]:checked').each(function() {
+		$(this).removeAttr('checked');
+	});
+
+		$('#group_tools_mail_member_selection input[name="select_all"]:checked').each(function() {
```
Phanoix: Is $('#group_tools_mail_member_selection input[name="select_all"]:checked').each(function() {... supposed to be duplicated?

PR #888, opened 2017-03-29
Modal accessibility changes

### LINKED ISSUE #844: Modal Accessibility
Modals that need to be modified for accessibility: ajax lightbox, profile edit, GEDS, opportunities platform opt-in, and onboarding.

Things that need to be done
  - Focus on modals
  - Have correct aria tags on each modal
  - Try using WET4 modals (Have js error on load)

ethanWallace:
Closes #844

### DIFF
--- a/mod/gc_onboard/views/default/welcome-steps/wire_modal.php
+++ b/mod/gc_onboard/views/default/welcome-steps/wire_modal.php
@@ -49,7 +49,7 @@
 </div>
 
 
-<div class="wb-overlay modal-content overlay-def wb-popup-mid" id="showWire" tabindex="-1" aria-live="assertive">
+<div class="wb-overlay modal-content overlay-def wb-popup-mid" id="showWire" tabindex="-1">
     <div class="">
         <div class="" id="welcome-step">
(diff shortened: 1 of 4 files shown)


mod/gc_onboard/views/default/welcome-steps/wire_modal.php
```diff
@@ -32,12 +32,24 @@
...
+						});
+                    });
+                </script>
+
+                      <?php }?>
 </div>
 
 
-<section class="wb-overlay modal-content overlay-def wb-popup-mid" id="showWire">
+<div class="wb-overlay modal-content overlay-def wb-popup-mid" id="showWire" tabindex="-1" aria-live="assertive">
```
nschonni: I'm not too familiar with Elgg, but I'm not seeing where dynamic content is coming into this section that would need a live region at all. Also, if this adds multiple live regions to the page, you're going to need code to syncronize them

piet0024: Good point. There is no dynamic content in this modal. I will remove the aria-live attribute.

Issue #874, opened 2017-03-27
Opportunity Platform creator selection error where multiple email addresses exist


ToferC:
The Opportunity Platform identifies an opportunity creator based on an email address, which can be duplicated for certain accounts (i.e., admin accounts).

In two cases, even though the account was created by a user, because the admin account has the same email, the OP creator was identified as the admin account.

A fix may include changing Opportunity Creator to a user.id.  Low priority as only applies to admins.

@smellems - I think this is what happened with you.

Steph4104:
Can I have a screenshot or a link where we can see the problem? Because I can't replicate the problem on my local version and I take a look a the opportunity of Sebastien and everyting look fine.

ToferC:
I tested thisbagain today and again had an opportunity created by admin.Chris. 

I'll stop by and we can walk through it together.

ToferC:
Hi @Steph4104 - I created a new opportunity here: https://gcconnex.gc.ca/missions/view/24960976 - same issue.

PR #898, opened 2017-04-03
Get thr right creator of opportunity 

Steph4104:
Get user from owner guid instead of account
#874

### DIFF
--- a/mod/missions/views/default/page/elements/mission-manager-info.php
+++ b/mod/missions/views/default/page/elements/mission-manager-info.php
@@ -13,9 +13,22 @@
 $mission = $vars['mission'];
 $container_class = $vars['container_class'];
 $grid_number = $vars['grid_number'];
-$test = $mission->account;
-$manager_account = get_user($mission->account); //Nick changed to owner_guid then back to account
-error_log('guid'. $manager_account->guid);
+$email = $mission->email;
+
+$accounts = get_user_by_email($email);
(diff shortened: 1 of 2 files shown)


mod/missions/views/default/page/elements/mission-manager-info.php
```diff
@@ -14,7 +14,7 @@
 $container_class = $vars['container_class'];
 $grid_number = $vars['grid_number'];
 $test = $mission->account;
-$manager_account = get_user($mission->account); //Nick changed to owner_guid then back to account
+$manager_account = get_user($mission->owner_guid); //Nick changed to owner_guid then back to account
```
piet0024: I remember the issue this caused by using guid (I just noticed my comment). Sometimes users will create a mission on behalf of their manager. So when they create the opportunity they want the contact person to be the manager and this is found by the email input I believe. So it will display the manager as the opportunity owner and not the user who filled out the form.

When we use owner_guid it will get the guid of the creator and they might not be the actual opportunity owner,

Perhaps a test will have to be done when creating the opportunity so that the account metadata contains the guid of the proper opportunity owner based on the email the user put in, and if the email is the same, default to the creator guid.

Steph4104: So, I think I fix it. What I did is I compare the email and the name, if I have a match, I take the guid from this

PR #899, opened 2017-04-03
Post notification fixes

pandurx:
micro mission empty body notifications, and removing ext.gcconnex.gc.ca


pandurx:
this was urgent as it affects all our users... this is currently now live on production

Issue #564, opened 2016-12-05
Allow for declining to add a Dates under Education in user profile


KevinChagnon:
As participation is voluntary and users are free to enter as much or as little information as they want (name and email at a minimum);
Please allow for users to decline to add Dates under "Education" in their user profile.
(allowing users to save their Education information without entering Start and End dates.

KevinChagnon:
+1

marc-donofrio:
New enhancements on hold due to maintenance mode.

PR #944, opened 2017-04-07
Allow no dates in education

claudielarouche:
Fixes issue https://github.com/gctools-outilsgc/gcconnex/issues/564

 I also added an empty row in the month selection as the month might also be unknown. I removed the comma between month and year because it didn't look good if both the year and month is missing.

I tested adding no end year, no start year, and both no end year and no start year. The check that start year is BEFORE end year also still works.

When there is no year information, it prints Unknown in English and Inconnu in French

I am not familiar with the possible ramifications with Missions so this may need to be tested.

education - fr

education - en

### DIFF
--- a/mod/b_extended_profile/languages/en.php
+++ b/mod/b_extended_profile/languages/en.php
@@ -51,6 +51,7 @@ $english = array(
     'gcconnex_profile:cancel' => 'Cancel',
     'gcconnex_profile:save' => 'Save',
     'gcconnex_profile:present' => 'Present', // for work and education time ranges.. from XX date to Present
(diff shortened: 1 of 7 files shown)


ToferC:
Thanks for this @claudielaroche. Suggest we might want to give users the option to not have a date. The use case was a concern of ageism - where users might not be selected for work because they graduated, for example, in 1984.  Unknown - Unknown raises the same kind of flags.

Suggest re-work with option being to either enter start & end dates or just omit the dates.

Thanks!

Phanoix:
Closing this for now since it a bit outdated now, some files changed dont's exist anymore, and it seems like it will need to be done somewhat differently.

Issue #426, opened 2016-11-04
Use Composer for the installation of Elgg and all the third party plugins


smellems:
This removes the burden of maintaining the code in our Repo.

Suggestion from Elgg team:
Separate your custom development into separate repos. Just treat them like a third party plugin. This way there is 1 master repo (your distribution), that is the repo you use to install/update (take a look at https://github.com/Elgg/starter-project. Plugin maintenance should be taken place in all the different plugin repos.

https://elgg.org/discussion/view/2570541/elgg-and-gcconnex-collaboration

nschonni:
This would probably be a good project for the hackathon

smellems:
I started something a few months ago to show how it would look: https://github.com/smellems/gcconnex-starter

If you look at the composer.json file, you can see most of the third party plugins used on GCconnex and the packages under smellems are ones created by the GCTools team or not available on Packagist

nschonni:
OK, so the ones under your account are ones you've manually pulled out from this repo? 
If so, could they get transferred to the org and published to Packagist, or is there a permission issue there?

nschonni:
@smellems actually it looks like filter branches could be used to extract the history of those mods out to their own repos https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/

smellems:
Yes they should all be under the gctools-outilsgc org.  I just did it in mine to test it out and show it.  Those repos are not up to date with the mods from gcconnex.  And keeping the history would be great!

For permissions, your right somebody from the GCTools would need to create all the repos and publish / sync them with Packagist.

nschonni:
Just tested it out with one and it seems to work well https://github.com/nschonni/GC_profileStrength
1. Checkout the gcconnex branch
1. Create a new branch with PLUGIN_NAME
1. Create the filter branch (this will wipe out the current branch and replace it) `git filter-branch --prune-empty --subdirectory-filter mod/PLUGIN_NAME/ PLUGIN_NAME`
1. Create a new empty GH repo with no license or README
1. Add the new remote `git remote add PLUGIN_NAME https://github.com/nschonni/PLUGIN_NAME.git`
1. Push the filtered branch to the empty repo `git push PLUGIN_NAME HEAD:master`

@smellems I guess then you can do what you you did with the composer file and add the new entry even without the Packagist publish

nschonni:
The IRCC team is working on this. We can discuss the PR and migration of the repos at the end of the Hackathon

[...]

PR #956, opened 2017-04-10
Initial Move to Composer

nschonni:
From the IRCC team at the hackathon for #426


nschonni:
Closing in favour of #973 for now since this is already out of sync with the plug-in folders. If that lands then individual PRs can be sent for the extractions

Issue #834, opened 2017-03-06
Inviting vs. Adding colleagues


joymosk:
2 issues: 
Language issue: The language is horrible. "Invite my colleagues to join you" makes no sense at all. I recommend we move to: To invite people you know to join you on GCcollab, please…. 
Calling it “Invite my colleagues” is very confusing as it isn’t clear if they are generic colleagues or GCconnex colleagues. Either Invite to join GCconnex
![invite my colleagues - gcconnex](https://cloud.githubusercontent.com/assets/24227707/23619532/f265c658-0261-11e7-99a9-f3bf753b9899.JPG)
 or Invite to be my GCcollab colleague

Usability issue: If they invite someone whose email is already on the system, it isn’t clear what they are to do next. 
Either more instructions in the error message, or a new functionality on that screen (“Request to Add colleague”) or this could be put under Colleague Requests.

joymosk:
Many of the instances of "my colleagues" have to be changed to "colleagues" 
See attached
![joy moskovic s my colleagues](https://cloud.githubusercontent.com/assets/24227707/23763336/1e9074be-04c7-11e7-8422-03f56e4e761f.JPG)

joymosk:
No the problem still exists. When is a correction going into prod?

PR #965, opened 2017-04-11
colleagues language update

Troy-Lawson:
language change. fixes issue #834

### DIFF
--- a/mod/rename_friends/languages/fr.php
+++ b/mod/rename_friends/languages/fr.php
@@ -177,9 +177,9 @@ $french = array(
 		
 // Invite Friends		
 		
-'friends:invite' => "Inviter des Colleagués",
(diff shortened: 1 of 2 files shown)


mod/rename_friends/languages/fr.php
```diff
@@ -177,9 +177,9 @@
 		
 // Invite Friends
```
Steph4104: It's "collègues" the translation of colleagues

Troy-Lawson: thanks! Maybe this year will be the year i finally do my french training....

Troy-Lawson: @Steph4104 the change has been made. Thanks for pointing it out

PR #973, opened 2017-04-12
Minimal move to Composer 

nschonni:
This is a minimized version of #956 so it just removes the composer files from source control, but without the plug-in extraction.

This is so stuff like #963 can be added through composer going forward, and tackle the custom plug-in extraction separately if that's easier for the devs.


nschonni:
I'll do a similar minimal PR to GCCollab if this one lands

nschonni:
Just wondering what the procedure is for landing this now that this has been approved?

nschonni:
Thanks! we'll try and look at smaller PRs to pull out the third party plug-ins next

Phanoix:
Should probably start with the ones we only changed some language in (should be a few of those) and see if we can get those accepted upstream.

smellems:
Great!  Happy this was merged!  Thanks Nick!

I started looking at 3rd party plugins and made a branch with one commit per plugin
https://github.com/smellems/gcconnex/commits/comp

I started doing some pull requests to upstream for language this morning, we'll see if they get merged.  For some of them it looks like even if they have a version that works for 1.12, it's not supported anymore (there is no branch)

lucbelliveau:
I just made a change to the missions plugins (#773) ; didn't realize this effort was underway.

I don't see a repo for the individual plugins under gctools; have they been created?  (If so where are they?)

nschonni:
I don't think they've split them yet. During the Hackathon the team from my department slit out the duplicate ones, but I parked that PR since it would have been harder to land and keep in sync.
Might be easier to create issues asking for repos to be create/migrated individually and then cut over

Issue #930, opened 2017-04-07
Remove the login dropdown in the header and replace with links to the login page


piet0024:
This dropdown causes issues on the login page for screen readers as the inputs and labels are duplicated. When trying to open this dropdown it also does not focus on the input.

Screen reader clicks "Link login" thinking it will load a new page but nothing happens as the dropdown opens. There is currently no feedback for screen readers, but if this was a link to the login page both sighted users and users using a screen reader can login to the site.

@CoolColin

piet0024:
This is also an issue for GCcollab

Issue #927, opened 2017-04-07
Remove additional login and register links in the footer when not logged in.


piet0024:
When using the "skip to about this site" feature when not logged in these links are focused first and not the actual about content. They are also redundant as they appear in the header.

@CoolColin

PR #977, opened 2017-04-13
Remove dropdown and add the link

Steph4104:
I put in comment all the dropdown code and add the link for the login page

#930 

Also, remove the second login in the footer #927

### DIFF
--- a/mod/wet4/views/default/page/elements/login_menu.php
+++ b/mod/wet4/views/default/page/elements/login_menu.php
@@ -210,7 +210,13 @@ $(document).ready(function () {
         $('#session').removeClass('active');
     });
 
-    $('#login_focus').click(function(){
+$('#signin-link').live('keydown', function(e) { 
+    if (keyCode == 13) { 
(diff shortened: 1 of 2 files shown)


mod/wet4/views/default/page/elements/login_menu.php
```diff
@@ -195,18 +195,24 @@
...
+    $('#signin-dropdown').click(function(e) {
+        e.stopPropagation();
+    });
+
+    $(document).click(function() {
+        $('#signin-dropdown').hide();
+        $('#session').removeClass('active');
+    });
+
+    $('#login_focus').click(function(){
```
piet0024: When I tab to the login and hit enter it does not perform this focus (It does when I click with the mouse though). Perhaps try adding this focus function to the link $('#signin-link')

Steph4104: @piet0024  I add a focus when user press ENTER on the login link. Also, I change the focus function to the id signin-link

Issue #919, opened 2017-04-06
Open message and notification drop down does not focus to element.


piet0024:
When a user tabs to messages or notifications on the user menu a link appears that can be tabbed to to open a drop down to give a preview of the user's messages. This link does not focus on the box. 

@CoolColin

piet0024:
This is also an issue for GCcollab

PR #981, opened 2017-04-19
Focus is on the first item in the dropdown

Steph4104:
Add focus to item in dropdown so screen reader will read the first line when the dropdown is open

#919

### DIFF
--- a/mod/wet4/views/default/ajax/notif_dd.php
+++ b/mod/wet4/views/default/ajax/notif_dd.php
@@ -7,7 +7,11 @@ This will populate a drop down that will show up when mousing over "notification
 
 @author: Nick
 */
-
+?>
+<script>
+$( "div.message-dd-holder div.message-dd-block a" ).first().focus(); //add focus on the first link
+</script>
+<?php
 $info_type = get_input('type');
 $info_count = get_input('count');
 $user = elgg_get_logged_in_user_guid();
--- a/mod/wet4/views/default/ajax/notif_dd.php
+++ b/mod/wet4/views/default/ajax/notif_dd.php
@@ -9,7 +9,10 @@ This will populate a drop down that will show up when mousing over "notification
 */
 ?>
 <script>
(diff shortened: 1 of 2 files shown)


Steph4104:
When I use the tab on notification, I got this. @ethanWallace 

image

Steph4104:
@ethanWallace I add a new focus. Can you test with a screen reader if it works ?

mod/wet4/views/default/ajax/notif_dd.php
```diff
@@ -7,7 +7,14 @@
 
 @author: Nick
 */
+?>
+<script>
+$( "li.elgg-menu-item-notifications div.message-dd-holder div.message-dd-block a" ).first().focus(); //add focus on the first link of notifications
+
+$( "li.elgg-menu-item-messages div.message-dd-holder div.message-dd-block a" ).first().focus(); //add focus on the first link of messages
```
nschonni: Setting a second focus will make the previous focus unnecessary. Which should be given focus?

mod/wet4/views/default/page/elements/message_dd_list.php
```diff
@@ -20,7 +20,7 @@
 		echo $no_results();
 		return;
 	}
-	echo "<p class='elgg-no-results message-dd-no-results'>$no_results</p>";
+	echo "<p class='elgg-no-results message-dd-no-results' aria-live='assertive'>$no_results</p>";
```
nschonni: `aria-live` is to indicate that there is dynamic content that will change. This doesn't make sense for a paragraph with static results

Steph4104: This paragraphe is in a dropdown menu generate by a ajax call. If aria-live is not there, the screen reader does not read the paragraphe.

mod/wet4/views/default/ajax/notif_dd.php
Steph4104: This focus is on a element from the dropdown in a ajax call. Only one focus is use at the time.

PR #990, opened 2017-04-20
TEAM NOP: Group Validation

kufii:
Adds form validation to groups


Phanoix:
Included as part of a broader form validation project.

PR #1014, opened 2017-04-26
gcforum patches and update

pandurx:


mod/gcforums/lib/functions.php
```diff
@@ -0,0 +1 @@
+<?php
```
nschonni: Is there something missing from here, or was it included by accident?

pandurx: I was in the middle of cleaning up (file structure and code) the plugin, but had to switch over to a different issue with a higher priority to fix.

pandurx:
@Steph4104 The text has been corrected and modified.

Issue #916, opened 2017-04-06
All Pages main content areas should follow the standardised semantic approach to headings h1-h6 


piet0024:
WCAG 2.4.6 Headings and labels
WCAG 1.3.1 Information and relationships

Starting at <main> each page must contain an h1 and the headers must make sense. (no more than 1 h1)

h1 - Welcome to Dashboard
h2 - Blog Widget
   h3 - Title of blog 
   h4 - Comments on blog
h2 - Colleague Widget

@CoolColin

piet0024:
This is also an issue that would need to be worked on for GCcollab

piet0024:
Groups issues: when you go to a blog or discussion in a group there are 2 h1s on the page. (the group name and the discussion title) when it should look like

h1 - Group title
h2 - Group content menu
h2 - Title of the group content
h3 - Comments

PR #1015, opened 2017-04-26
Opportunity platform heading structure accessibilty

Phanoix:
Part of #916


mod/missions/views/default/page/elements/main-members.php
```diff
@@ -5,6 +5,10 @@
  *
  * License: Creative Commons Attribution 3.0 Unported License
  * Copyright: Her Majesty the Queen in Right of Canada, 2015
+ *
+ *
+ * MODIFICATION
+ * Nick - Modifed search result list / Modified headings
```
nschonni: This is what source control is for rather than adding inline history comments

Phanoix: That was actually merged as part of another PR.

PR #1017, opened 2017-04-27
Upgrade to latest version of PHPMailer 5.2.23

smellems:


nschonni:
If you don't have customization in this module, it might be a good time to just flip to Composer to do the upgrade

smellems:
I didn't find a newer version of the Elgg PHPMailer plugin.  That's why I just updated it here.  Not sure if we could just include PHPMailer (not the Elgg plugin) using composer.

https://packagist.org/packages/phpmailer/phpmailer

nschonni:
Ah, the plug-in hasn't been touched in awhile https://github.com/cash/elgg-phpmailer and never got Composer'd
Not sure if deleting the `vendors` and installing it at the project level would work or not

PR #1019, opened 2017-04-27
Tab to add fil from embed content in IE

### LINKED ISSUE #1018: Tab in embed content in IE
You can't add a file using the tab in embed content if you are on IE. You are able to do all the step but the file did not add to the ckeditor.

If you are using the mouse, everything is fine.

BUT, if you try using the mouse first and add a file, and after without refreshing the page, try with the tabs, it works.

Steph4104:
Remove lightbox.php from js folder in the wet4 theme

fixes #1018


Steph4104:
The file in the wet4 is the same in elgg. 
 So, I'm not sure what this file did exactly, and what can be affected by this since there are no change.

PR #1023, opened 2017-04-28
Remove error from ajax call

### LINKED ISSUE #1009: Unexpected error when performing an AJAX call. Maybe the connection to the server is lost.
This error seems to appear when you leave GCconnex open for a long time (in this case over night) and return to the open window.  
Please remove this error message as the message provides no useful information to the user but makes the site look "broken" when it is not: the user can continue to use the site without consequence.

ajax

Steph4104:
add file to wet4 plugin and comment the error message

Fixes #1009


js/lib/ajax.js
```diff
@@ -102,7 +102,7 @@ elgg.ajax.handleOptions = function(url, options) {
  * @private
  */
 elgg.ajax.handleAjaxError = function(xhr, status, error) {
-	elgg.register_error(elgg.echo('ajax:error'));
+	//elgg.register_error(elgg.echo('ajax:error'));
```
nschonni: Maybe just delete the handler entirely?

PR #1036, opened 2017-05-02
Adding skills to opp card

piet0024:
Modifying view of the opportunity cards to add the users skills.

There are different pages / views that handle member lists and some of these don't grab the "why this user was found" information. 

Added 4 skills to the card and a link to open a modal to view all the user's skills.

### DIFF
--- a/mod/missions/views/default/user/mission-candidate.php
+++ b/mod/missions/views/default/user/mission-candidate.php
@@ -60,10 +60,20 @@ $user_skills = elgg_get_entities($options);
 
 $skill_set = '';
 $count = 1;
-foreach($user_skills as $skill) {
-	$skill_set .= '<span name="user-skill-' . $count . '" class="mission-skills">' . $skill->title . '</span>';
+//Limiting how many skills first show up in the array
+$limit_skills = array_slice($user_skills, 0, 4);
+foreach($limit_skills as $skill) {
+    
+	$skill_set .= '<span name="user-skill-' . $count . '" class="mission-skills cut-skill" title="'.$skill->title.'">' . $skill->title . '</span>';
 	$count++;
 }
+
+foreach($user_skills as $skill) {
+    
+	$skill_set_more .= '<span name="user-skill-' . $count . '" class="mission-skills" title="'.$skill->title.'">' . $skill->title . '</span>';
+	$count++;
+}
+
 
 // Displays invitation button if the user is opted in to micro missions.
 $button_content = '';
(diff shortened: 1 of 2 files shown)


mod/missions/views/default/user/mission-candidate.php
```diff
@@ -143,21 +143,22 @@
...
         <div>
             <?php echo $skill_set; ?>
-        </div>
-
-	</div>
-    </div>
-    <div class="user-button-content">
-        <p><a href="#mid-screen-<?php echo $user->guid;?>" aria-controls="mid-screen" class="overlay-lnk center-block" role="button">View All Skills</a></p>
+        
+        <p class=""><span class="center-block"><a href="#mid-screen-<?php echo $user->guid;?>" aria-controls="mid-screen" class="overlay-lnk" role="button">View All Skills</a></span></p>
```
nschonni: Adding empty `class` on tag?

nschonni: I realize it was there before, but the `aria-controls` is pointing at an ID that doesn't exist. It looks like the generated ID of the element is `mid-screen-<?php echo $user->guid;?>`

Steph4104:
opt_card_design

mod/missions/views/default/user/mission-candidate.php
```diff
@@ -60,11 +60,21 @@
 
 $skill_set = '';
 $count = 1;
+//Limiting how many skills first show up in the array
+$limit_skills = array_slice($user_skills, 0, 4);
+foreach($limit_skills as $skill) {
+    
+	$skill_set .= '<span name="user-skill-' . $count . '" class="mission-skills cut-skill" title="'.$skill->title.'">' . $skill->title . '</span>';
+	$count++;
```
nschonni: Wouldn't the count be up to 4 greater than the actual size since you're counting the first 4 twice?

mod/missions/views/default/user/mission-candidate.php
```diff
@@ -60,10 +60,22 @@
 
 $skill_set = '';
 $count = 1;
-foreach($user_skills as $skill) {
```
nschonni: Not sure if this is correct PHP, but
```php
$skill_class = 'mission-skills cut-skill'
foreach($user_skills as $skill) {
	$skill_set .= '<span name="user-skill-' . $count . '" class="'. $skill_class .'" title="'.$skill->title.'">' . $skill->title . '</span>';
	$count++;
	if($count == 4) {
		$skill_class = 'mission-skills'
	}
}
```

PR #1042, opened 2017-05-04
resolved troubleshoot tool for notification

pandurx:
@KevinChagnon keeping you updated for this fix


pandurx:
tested on localhost, preproduction and made sure the records in the database were correctly removed when admin user unsubscribes a user from specified content in the  tool

PR #1049, opened 2017-05-04
Db connect fix for read/write split

### LINKED ISSUE #1041: Update direct references to $CONFIG->db with appropriate function calls
Direct references to $CONFIG->db causes problems with read/write db split, either use getConnectionConfig() and isDatabaseSplit() or have elgg create the connection object with getLink($type) for example.

Phanoix:
Some mods create their own connections to the DB, this ensures they user the appropriate db config variables in both regular and read/write split database configurations.
fixes #1041


nschonni:
Not sure if Memcache might be an easier built-in solution https://elgg.org/discussion/view/2654394/is-memcached-any-faster

PR #1068, opened 2017-05-09
script to fix all the forum-type subscriptions without group membership

pandurx:
since the forums work differently than the rest of the entities in gcconnex (because you can have nested forums, forums can contain forum topic, etc through recursive hierarchy in the database) developed a script to remove any forums that the user is not part of the group. 
LMK if there is a better way of querying against the database, I have it in separate parts to extract information and then determine if the user is a member of the group.


mod/cp_notifications/actions/cp_notifications/fix_forums_subscription.php
```diff
@@ -0,0 +1,110 @@
...
+$original_events = _elgg_services()->events;
+$original_hooks = _elgg_services()->hooks;
+_elgg_services()->events = new Elgg\EventsService();
+_elgg_services()->hooks = new Elgg\PluginHooksService();
+elgg_register_plugin_hook_handler('permissions_check', 'all', 'elgg_override_permissions');
+elgg_register_plugin_hook_handler('container_permissions_check', 'all', 'elgg_override_permissions');
+
+$insert_in_text = array();
+$str_id = elgg_get_metastring_id('fix_forum_subscriptions');
+$val_id = elgg_get_metastring_id('0');
```
Phanoix: Doesn't look like you use these anywhere

nschonni:
Not sure if this is a better approach http://learn.elgg.org/en/v2.3.2/guides/upgrading-data.html

pandurx:
@nschonni i'll look into it (though, it may not work as it's for v2.3.2 and our version is at v1.12).

nschonni:
Yeah, it looks like it is even further down the road (v3) https://github.com/Elgg/Elgg/pull/10118

mod/cp_notifications/actions/cp_notifications/fix_forums_subscription.php
pandurx: i have removed this

pandurx:
I'll probably upgrade to using the approach once we upgrade Elgg

PR #1071, opened 2017-05-10
Updated readme with instructions to install htaccess

lucbelliveau:
Updated the readme to indicate how to install the required .htaccess file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,10 @@ use docker-compose to start/create your containers.
 
     docker-compose up
 
+Install the `.htaccess` file in the root of the repo.
+
+    cp ./install/config/htaccess.dist ./.htaccess
+
 Then visit [http://localhost:8080](http://localhost:8080) and follow the
 instructions to complete your installation.  Once the installer is complete,
 refer to [INSTALL.md](INSTALL.md#configure-plugins) to configure the plugins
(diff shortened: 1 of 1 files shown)


nschonni:
Should that just be in the Dockerfile you added?
PS: thanks for doing the Docker stuff, it really helped our group for the hackathon

lucbelliveau:
Np, I'm glad it helped.

docker-compose mounts the project directory as the webroot to allow live editing; so it wouldn't be able to copy anything into that directory at build time (from the Dockerfile).... if we wanted to have it done automatically we could create a custom entrypoint script; but given the only thing we need to do is copy a file, I figured I would just add it to the readme.

smellems:
When you go through the installation, the .htaccess file is create in the root directory.  But Apache need to have access.  We could add a chmod to the Dockerfile in the CMD.

lucbelliveau:
So you just ignore the error about testing redirects?  (That's why I had placed the file there manually - it was telling me it couldn't test redirects)  If that's the case I would just mention it in the readme... 

Regarding CMD, those commands are executed at build time, and because we are not copying the files we are mounting them, it wouldn't persist...  unless you aren't using docker-compose?

smellems:
I haven't been using docker-compose, just docker.  I changed the Dockerfile for GCcollab https://github.com/gctools-outilsgc/gccollab/blob/gccollab/Dockerfile  At the bottom are the 3 docker commands to build and start the containers.

lucbelliveau:
The current docker-compose method that enables live editing will still require the user to manually copy the .htaccess file... because the root is mounted as a volume.

Given all the changes with composer, I'll wait and see what happens... I have a few ideas how to automate the installation and save live editing...  Once I see composer in action it will be clearer.  I will close this PR until then.

PR #1074, opened 2017-05-10
updated plugin order settings and added a generator for easier updates

lucbelliveau:
Updated plugin list in config file, and added a method to automatically generate config based on existing configuration when needed.


nschonni:
This is a third party plug-in so it probably shouldn't be modified https://github.com/lowfill/elgg_default_plugin_order

lucbelliveau:
That's a good point; but that project is dead... 

I could create a fork?

nschonni:
Don't know, the other thought might be to move the required logic to the/a install script https://github.com/gctools-outilsgc/gcconnex/blob/gcconnex/install/cli/sample_installer.php
But I'm not :100: that Elgg works that way :wink:

lucbelliveau:
Since it's not a 1 time thing, I decided to fork the plugin itself and add a import/export button to the admin interface.  Follow the instructions [here](https://github.com/lucbelliveau/plugin_loader/blob/master/README.md) for testing, or from INSTALL.md.

Issue #1057, opened 2017-05-08
Merge GCcollab back into GCconnex (Composer)


smellems:
Remove differences between GCconnex and GCcollab, either by moving customization to mods or merging changes from GCcollab into GCconnex.

The goal is to have a common "GCelgg" with custom platform specific mods being managed in separate repos ( gctools-outilsgc/gcconnex#1056 ).

markwooff:
Most of the shared modules that have been customized on the GCcollab side have been moved to their own module (e.g. `gcRegistration` to `gcRegistration_collab`, `gc_splash_page` to `gc_splash_page_collab`, etc.). The originals would then sit in this "GCelgg" repo while we maintain our GCcollab-specific modules.

smellems:
The `gcSplash` and `gcSplash_collab` should be two new repos.  The idea is that we would remove the GCcollab repo since it would be the same as GCconnex (GCelgg).  Then you can build one or the other using a different composer file

nschonni:
It may be a separate issue, but looking at whether the changes between the Collab and Connex plug-ins can just be moved to config instead of code might be a good idea.
EX: If the differences for the Splash are just image and maybe translation, that could just be config boxes on the Admin side

markwooff:
The differences between `gc_splash_page` and `gc_splash_page_collab` are small, it would make sense to have some sort of config to set the variables between the two platforms. In this case the differences are some links, text, and CSS.

@smellems is it possible to use composer to pull different data for a plugin? For example, passing through a `gcconnex` or `gccollab` flag to pull in the correct vars for each. If not, we could always create a small Elgg admin panel where you could choose between the two platforms.

nschonni:
Looks like it is covered in the dev docs http://learn.elgg.org/en/stable/guides/settings.html. Looks like you could probably use the installer.php like https://github.com/gctools-outilsgc/gcconnex/blob/gcconnex/install/cli/travis_installer.php to config the instance specific stuff

smellems:
As a first step, we will bring `gccollab` mods into the `gcconnex` repo.  The `gccollab` repo will become a copy of the `gcconnex` repo and all development will be on `gcconnex`.  There will be two plugin order config files to enable either `gcconnex `or `gccollab` mods.

Differences between core `gcconnex` and `gccollab` will be merged into `gcconnex` or moved to mods.

markwooff:
[...]

Issue #1056, opened 2017-05-08
Create separate Repos for mods specific to GCcollab or GCconnex (Composer)


smellems:
For all mods that are specific to GCcollab or GCconnex (ex.: Wet4), create seperate Repos and add them to Packagist for installation via Composer.

PR #1082, opened 2017-05-12
Build: Add Docker for Composer install

nschonni:
Realized with the removal of the previously checked in mod & vendor files, the Docker setup would be broken.
The mod and vendor folders are masked in the volume mount so they doen't get wiped out

/cc @lucbelliveau

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,10 @@ FROM ubuntu:14.04
 MAINTAINER Luc Belliveau <luc.belliveau@nrc-cnrc.gc.ca>
 
 # Install dependencies
+ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update
-RUN apt-get install -y git apache2 php5 libapache2-mod-php5 php5-mysql php5-gd php5-curl
+RUN apt-get install -y git apache2 php5 libapache2-mod-php5 php5-mysql php5-gd php5-curl curl
+RUN curl -sS https://getcomposer.org/installer | sudo php -- --install-dir=/usr/local/bin --filename=composer
 
 # Enable Apache mod_rewrite
 RUN a2enmod rewrite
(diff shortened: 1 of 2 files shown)


lucbelliveau:
Is the plan to install all the mods via composer?

docker-compose.yml
```diff
@@ -29,6 +29,8 @@ services:
     volumes:
       - .:/var/www/html
       - ./data/data:/data
+      - /var/www/html/mod
```
lucbelliveau: This will disable live edits of mods ... the mods will be loaded from the docker image rather than mounted from a host volume.. you would need to rebuild the docker image each time changes are made to modules.

nschonni:
> Is the plan to install all the mods via composer?

@lucbelliveau I think that is what some are pushing for #426, #1056, #1057

docker-compose.yml
nschonni: Your right, but I thought this was better that listing out all the Composer installed modules since that list is likely to grow

Dockerfile
```diff
@@ -20,8 +22,12 @@ RUN sed -i '/ErrorLog ${APACHE_LOG_DIR}\/error.log/c\ErrorLog \/dev\/stderr' /et
 # Uncomment to also see access log (rebuild with docker-compose up --build)
 # RUN sed -i '/CustomLog ${APACHE_LOG_DIR}\/access.log combined/c\CustomLog \/dev\/stdout combined' /etc/apache2/sites-available/000-default.conf
 
+COPY . /var/www/html
```
nschonni: @lucbelliveau you could use this for the `.htaccess` stuff in #1071

lucbelliveau:
Ok, overall I think its a step in the right direction.

Do you use Docker for development at all, or do the devs have their own local installs?

Editing modules could become a big pain though... needing to type "composer update" for even the slightest code change adds up to alot of time on complex projects like this... particularly if you are working on a feature that spans different modules.

Do you have anything in mind to manage that yet?

nschonni:
> Do you use Docker for development at all, or do the devs have their own local installs?

Not sure, i'm not actually on the GCTools team :wink: 



> Editing modules could become a big pain though... needing to type "composer update" for even the slightest code change adds up to alot of time on complex projects like this... particularly if you are working on a feature that spans different modules.

Yeah, I don't know if adding an additional volume mount through the CLI would make sense for that case
`docker-compose up --volume ./mod/foobar:/var/www/html/mod/foobar` (don't think that is a real CLI option)
When I was messing around with this I was just using `docker exec -i gcconnex_gcconnex sh -c "cd /var/www/html/ && composer require foobar"

PR #1093, opened 2017-05-17
Customize/style question mod for GCconnex

ethanWallace:
Customized questions mod to fit style and have consistent look to GCconnex. Also changes were made to make mod more user friendly to users with AT.


nschonni:
Is the plan to PR the changes to questions upstream? Note that we are a few versions behind them because of the Elgg 1.12 support, but we were able to get them to add our translations at least

ethanWallace:
@nschonni As mentioned above, these changes are mainly for integrating the questions mod nicely into GCconnex. A lot of the changes have to do with giving the questions a similar visual style to match other content types on our site. It is possible a few of these changes can be pushed to the original mod but we are not focusing on that at this time.

PR #1098, opened 2017-05-17
Exceeded number of characters for Notifications

pandurx:


nschonni:
The DOCX file should probably be rebased out so it doesn't bloat the repo

mod/cp_notifications/lib/functions.php
```diff
@@ -153,6 +153,7 @@ function isJson($string) {
 
 /**
  * assembles the digest then encodes the array into JSON to be saved
+ * EDIT (May 16 2017): change to new database structure
```
nschonni: Delete me?

mod/cp_notifications/start.php
```diff
@@ -327,15 +301,18 @@ function cp_overwrite_notification_hook($hook, $type, $value, $params) {
 
 
 		case 'cp_group_mail': // group_tools/actions/mail.php
+		error_log(">>>> poop!");
```
nschonni: ?

mod/cp_notifications/start.php
```diff
@@ -327,15 +301,18 @@ function cp_overwrite_notification_hook($hook, $type, $value, $params) {
...
 				'cp_group' => $params['cp_group'],
 				'cp_group_subject' => $params['cp_group_subject'],
 				'cp_group_message' => $params['cp_group_message'],
 				'cp_msg_type' => $cp_msg_type
 			);
 			$subject = elgg_echo('cp_notify:subject:group_mail',array($params['cp_group_subject'],$params['cp_group']['name']),'en') . ' | ' . elgg_echo('cp_notify:subject:group_mail',array($params['cp_group_subject'],$params['cp_group']['name']),'fr');
-			foreach ($params['cp_group_mail_users'] as $to_user)
+			foreach ($params['cp_group_mail_users'] as $to_user) {
 				$to_recipients[$to_user] = get_user($to_user);
+				error_log(">>>> poop! 2 : ".get_user($to_user)->email);
```
nschonni: ?

pandurx: oops. I'll have them removed.

mod/cp_notifications/lib/functions.php
```diff
@@ -373,30 +442,51 @@ function create_digest($invoked_by, $subtype, $entity, $send_to, $entity_url = '
...
 
-	// save the information to the digest object for later access
-	$ia = elgg_set_ignore_access(true);
-	$digest->description = json_encode($digest_collection);
-	$digest->save();
-	elgg_set_ignore_access($ia);
+	$group_name = base64_encode($group_name);
+
+	/// check if record exists already, if not then proceed.
+	$query = "SELECT 1 FROM notification_digest WHERE entity_guid = {$entity_guid} AND user_guid = {$user_guid} AND notification_entry = '{$notification_entry}'  LIMIT 1";
```
Phanoix: Adding an index over the columns in this query might not be a bad idea.

pandurx: yeah, i was thinking the same. probably user_guid and notification_entry?

Phanoix: I'd say all 3 to make it cover the whole query, the order would depend on what would be more useful for other queries that are run when the digests are sent out.

Phanoix: pretty sure user_guid should be first, so maybe (user_guid, notification_entry, entity_guid) ? I think that should cover everything actually.

pandurx: i can include the index in the script tomorrow.

pandurx: i can't have notification_entry as an index. http://stackoverflow.com/questions/1827063/mysql-error-key-specification-without-a-key-length

Phanoix: Right, it's a longtext... well, it'll still be faster than the old version.

PR #1126, opened 2017-05-31
Enhanced user search

### LINKED ISSUE #495: Find Members Search Query needs to be optimized
Searching from Find Candidates / Find Members takes 12+ seconds to resolve.  This needs to be faster for a good user experience.  Require heavy optimization of search or reduce scope to only search for skills and update search box text.

lucbelliveau:
Fixes gctools-outilsgc/gcconnex#495 

# What it does
This creates the discussed materialized view to speed up search operations.  It is isolated by introducing a new plugin.

# How to test
1. Enable the plugin
2. Wait half an hour or so, or visit http://siteurl/cron/halfhour/ to trigger the update immediately.
3. Perform a search

## Notes

The search index is currently set to update every half hour - this requires that Elgg's [cron feature](http://learn.elgg.org/en/2.0/admin/cron.html) is properly configured on the server.  (I noticed the example crontab file is erroneous and wouldn't work without modification - specifically crontab does not support variable expansion during variable assignments... )

It would be good to monitor the time it takes for the update to occur, you can generate the index manually by executing the following from mysql:

```sql
CALL NRC_EUS_Refresh_MemberSearch();
```

On my system the updates takes about 20 seconds.  During this time there is no service impact, so it is safe to run.  Also until the first refresh is completed, the search will fallback to the current method - so there is no service interruption at all.

It is also safe to turn the plugin off after it's been enabled - the search will fallback to the current implementation if the plugin is not enabled.

### DIFF
--- a/mod/enhanced_user_search/lib/sql.php
+++ b/mod/enhanced_user_search/lib/sql.php
@@ -1,0 +1,114 @@
+<?php
+
+namespace NRC\EUS\SQL;
+
+class Constants {
+  public function get() {
+    $dbprefix = elgg_get_config("dbprefix");
+
+    $tableName = "NRC_EUS_mvMemberSearch";
+    $procName = "NRC_EUS_Refresh_MemberSearch";
+
+    $BUILD_SQL = "
+    CREATE TABLE IF NOT EXISTS `{$tableName}` (
+      user_guid bigint(20) unsigned not null,
+      name text not null,
+      contactemail text,
+      education text,
+      gc_skills text,
+      portfolio text,
+      work text
+    );";
+
+    $INDEX_SQL = "
+    CREATE FULLTEXT INDEX NRC_EUS_MemberSearch ON
+      {$tableName}(
+        name,
+        contactemail,
+        education,
+        gc_skills,
+        portfolio,
+        work
+      );";
+    $REFRESH_PROC = "
+    CREATE PROCEDURE `{$procName}` ()
+    BEGIN
+      TRUNCATE TABLE {$tableName};
+
+      INSERT INTO {$tableName}
+      SELECT
+        a.guid as user_guid,
+        a.name as name,
+        c.string as contactemail,
+        GROUP_CONCAT(DISTINCT f1.title SEPARATOR ', ') as education,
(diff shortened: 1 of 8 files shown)


ToferC:
Thanks Ilia. Is this live on pre-prod?

Phanoix:
It is but hasn't been activated yet.

ToferC:
@Phanoix - please advise when this is on pre-prod for testing. Thanks!

Phanoix:
finished setting it up on preprod, tried to run the update and get this error: Row 1159 was cut by GROUP_CONCAT()

mod/enhanced_user_search/lib/sql.php
```diff
@@ -0,0 +1,156 @@
...
+
+      INSERT INTO `{$tableName}_tmp`
+      SELECT
+        a.guid as user_guid,
+        a.name as name,
+        c.string as contactemail,
+        GROUP_CONCAT(DISTINCT f1.title SEPARATOR ', ') as education,
+        GROUP_CONCAT(DISTINCT f2.title SEPARATOR ', ') as gc_skills,
+        GROUP_CONCAT(DISTINCT f3.title SEPARATOR ', ') as portfolio,
+        GROUP_CONCAT(DISTINCT f4.title SEPARATOR ', ') as work
```
Phanoix: @lucbelliveau  One of these group_concat seems to be erroring out when I try to run the update on preprod with "Row 1159 was cut by GROUP_CONCAT()".

lucbelliveau:
@Phanoix From the mysql console, can you execute the following:

    DROP procedure NRC_EUS_Refresh_MemberSearch;

Then just reload any page on the site... it should re-create the procedure... and the group_concat error should be resolved.

Phanoix:
Thanks, looks like 15000 is long enough. The update ran a little over 2 minutes.
@ToferC it's ready for testing.

lucbelliveau:
@Phanoix Thanks!  Do you already have cron setup on the servers?

Phanoix:
Some of them should be running, not sure about the halfhour one, will have to check with @cdegrass .

ToferC:
Looks good here. Approved for move to prod - but let's watch carefully. Maybe turn on in evening and plan some live tests.

PR #1128, opened 2017-06-02
increased opportunity description limit from 8000 to 12000

Phanoix:


ToferC:
@Phanoix - I think this is done now.  Please close if so.

Issue #322, opened 2016-10-28
multi post messages when someone puts something new on GCconnex


Troy-Lawson:
GCCON-178
From: Marc.Turcotte@nserc-crsng.gc.ca 
 Just now I just got 7 emails because someone posted a new item to one of the groups I follow. 
 
 The link in for one of the emails is this: https://gcconnex.gc.ca/messages/read/21393536

Phanoix:
Do the items also end up posted multiple times and/or appear multiple times under activity?

pandurx:
It does.. most of the time (with different users)

Phanoix:
So the rest of it (multiple notifications, activity entries) is working as intended considering the content ends up created multiple times, and we just need to prevent that from happening... I'll check if anything has been done recently regarding this in the newer elgg version.

Phanoix:
Confirmed to occur as a result of database load spikes / long table locks.

sboots:
Hey team! Quick note that I'm also seeing this behaviour when posts are made, or bookmarks are added, to large groups (1000+ members). 

Typical steps to replicate are,
1. find a large group (eg. https://gcconnex.gc.ca/bookmarks/group/18885366/all)
2. add a new bookmark
3. after clicking "Add a bookmark" on the form to save it, the browser will stall while waiting to load the page
4. depending on the browser, it may try to re-submit the form automatically, causing several duplicate posts to appear.

One possibility is that this is related to the email notifications that are sent out to every member of the group (if they're done synchronously with the page load rather than asynchronously in some other process). That might explain why this is only an issue with really large groups.

Thanks and rock on! 

![gcconnex-multi-post-20170118](https://cloud.githubusercontent.com/assets/263472/22068419/f69e2cb6-dd62-11e6-8cb8-d212cee4c829.png)

ToferC:
This is super useful context @sboots - thanks!

pandurx:
See tbs-sct/gcconnex#351

Troy-Lawson:
same as gctools-outilsgc/gcconnex#351

ToferC:
Closing as issue addressed by @Phanoix .

PR #1141, opened 2017-06-06
use extra unique entity column to prevent duplicate objects - attempt 1 (weaker constraint)

Phanoix:
Prevents duplicate objects through the use of a unique column based on title,discription, and ownerGUID and timestamp

Duplicate objects that are submitted at different times to the server will not be filtered out with this, if we get any more after this is implemented I can add stronger constraints.

requires adding extra column to the entities table:
```` sql
ALTER TABLE {prefix}entities ADD COLUMN `duplicate_check` char(100) UNIQUE NULL 
````
This way older entries will remain null and anything that's not an object will be unaffected, but duplicates will be prevented from being added to the database on the fist step.

If this works, I'll add this column to the install scripts.

### DIFF
--- a/engine/classes/ElggObject.php
+++ b/engine/classes/ElggObject.php
@@ -146,7 +146,8 @@ class ElggObject extends \ElggEntity {
 			// TODO(evan): Throw an exception here?
 			return false;
 		}
-		
+
+		$this->enable();
 		return $guid;
 	}
 
--- a/engine/classes/ElggEntity.php
+++ b/engine/classes/ElggEntity.php
@@ -1662,13 +1662,24 @@ abstract class ElggEntity extends \ElggData implements
 				return false;
 			}
 		}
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by Phanoix** [comment]
The travis checks will fail until I add the column to the install script it's using.

File: engine/classes/ElggObject.php [hunk 1]
```diff
@@ -146,7 +146,8 @@ protected function create() {
 			// TODO(evan): Throw an exception here?
 			return false;
 		}
-		
+
+		$this->enable();
```
nschonni: Is the change to the Elgg core needed or could it be done somewhere else?

Phanoix: it can't really go anywhere else, but hopefully it's only temporary.

pandurx:
relates to issue #322

PR #1144, opened 2017-06-07
Tag plugin work in progress

### LINKED ISSUE #1108: Tagging interface enhancement

piet0024:
First version of Tags and community plugin. Currently only works for creating a blog.

This is still a work in progress. Feedback about how it looks or works is welcome at this point.

Should be on our preprod environment.

Closes #1108


piet0024:
Things to do:
-Modify the remove all option on the community input / make it text / change location
-When selecting All Public servants, user can't choose another community
-Save the community of practice array somewhere (in db, plugin setting, JSON)
-Have an interface for site admin to add or remove to the community of practice array
-Hook into object creation actions instead of overriding them
-Extend creation forms rather than overriding them
-Helper / FAQ pages for what communities are and how they work
-Modify form view for editing content (perhaps we can keep the same modal view)
-Extend the views and actions for all other content types other than blogs

joymosk:
Could someone explain how this works? @piet0024 or @Phanoix

PR #1152, opened 2017-06-13
Display similar questions on question creation

ethanWallace:
Runs when a question is first being created to avoid duplicating questions on the platform.


mod/wet4/actions/object/question/autocomplete.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
piet0024: Do we want to put this action in the questions mod if we are going to maintain it.

ethanWallace: The main reason I did it in our mod is to deal with features that we have(French/English content). At a later time we can go back and push this to the questions mod without our unique features in it.

PR #1177, opened 2017-06-20
System message changes

### LINKED ISSUE #1110: Ability to close system message with keyboard
When a system message appears from a page load (not an ajax system message/ ex: 'Welcome to GCconnex %s' on login) users cannot close the message with their keyboard.

ethanWallace:
Can now close system messages by keyboard and by clicking anywhere on the message. Also changed CSS so the message stays the width of the window.

Fixes #1110

### DIFF
--- a/mod/wet4/views/default/wet4_theme/custom_css.php
+++ b/mod/wet4/views/default/wet4_theme/custom_css.php
@@ -1385,7 +1385,7 @@ border:none;
     width: 95%;
     max-width: 1140px;
     z-index: 1039;
-    top: 185px;
+    top: 85px;
 }
 
 .elgg-state-success {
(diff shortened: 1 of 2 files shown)


mod/wet4/views/default/wet4_theme/custom_css.php
```diff
@@ -1382,8 +1382,9 @@
 
 .custom-message{
     position: fixed;
-    width: 1140px;
-    z-index: 9999;
+    width: 95%;
+    max-width: 1140px;
+    z-index: 1039;
     top: 185px;
```
piet0024: can we modify the top a little so that it's not in the way of the site content? I know they want it to fade away but I don't know about the accessibility of that. Tabbing to the X works and clicking anywhere on the alert also dismisses it 👌

PR #1222, opened 2017-07-06
small forum (edit topic) change

pandurx:


pandurx:
@Phanoix it is a very minor change

Issue #1223, opened 2017-07-06
Search for parent group when making a group a subgroup of another group is not showing results


KevinChagnon:
When searching for a group using the "make this group a sub-group" feature, you have to include the "language code" in the group name otherwise there will be no results.

To find group "INAC Alberta" you must search for {"en":"INAC Alberta","fr":"AANC Alberta"}

Link to group: https://gcconnex.gc.ca/groups/profile/26794819/eninac-albertafraanc-alberta

This way of displaying the group name is also seen within the group settings under "Manage group operators"

PR #1226, opened 2017-07-07
Search to add a subgroup + group operator title+ welcome message

Steph4104:
Few language fix.

The search for a group to make it a subgroup now search in all the string and not the start of the string.
Title of the group operators page is fix.
Fix issue #1223 
The name of the group in the exemple of the group welcome message is fix.

### DIFF
--- a/mod/au_subgroups/views/default/au_subgroups/search_results.php
+++ b/mod/au_subgroups/views/default/au_subgroups/search_results.php
@@ -24,7 +24,7 @@ if (empty($vars['q'])) {
 }
 else {
   $query = sanitize_string($vars['q']);
-  $options['wheres'][] = "g.name LIKE '{$query}%'";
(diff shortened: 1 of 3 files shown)


mod/au_subgroups/views/default/au_subgroups/search_results.php
```diff
@@ -24,7 +24,7 @@
 }
 else {
   $query = sanitize_string($vars['q']);
-  $options['wheres'][] = "g.name LIKE '{$query}%'";
+  $options['wheres'][] = "g.name LIKE '%{$query}%'";
```
Phanoix: That's a much slower search, we might have to maybe have it do
```php
 g.name LIKE '{\"en\":\"{$query}%'";
```
if this is too slow on prod

Steph4104: If we want to do something like that, we need to think when someone look for something in french...

Issue #1221, opened 2017-07-06
Links to any Ted.com video put in iframe causes an immidiate redirect to the video's URL when accessing the group page


KevinChagnon:
As a group owner I would like the page of videos of interest on Ted.com in the iframe widget with it causing a redirect of my group page to the video on Ted.com.

One the URL to ted.com is entered in the widget settings and saved, the only way to stop the redirect is to click on the "X" in the by the browser's address bar to stop the page from loading.  If lucky enough, the widget will have loaded on the page without doing the redirect allowing you to delete it but you have to be quick!

I could not replicate the behaviour using the URL to another website in iframe settings.

PR #1228, opened 2017-07-07
Prevent for redirect iframe

Steph4104:
Add sandbox option on ifram to block top navigation

issue #1221


mod/wet4/views/default/output/iframe.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * Display a page in an embedded window
+ *
+ * @package Elgg
```
nschonni: This doesn't look right

Issue #1193, opened 2017-06-27
Contact Form resets the reason drop down when returning with a validation error


KevinChagnon:
The form resets the Reason drop-down when returning with a validation error - As a regular user, I do not want to have to re-enter my information - an indication of the field that should be corrected would be sufficient.

Not selecting a Reason also returns the "file type validation" when no file was attached; 

There is a number one ( 1 ) at the top of the Contact Form that should not be there;

![valid file type](https://user-images.githubusercontent.com/23527358/27596500-bbde341a-5b2d-11e7-9976-8e8c7884fcd9.jpg)

ToferC:
Approved for work.

PR #1238, opened 2017-07-10
Don't reset value in the contactform

Steph4104:
Remove 1
Keep the value when user have an error in the validation of the contactform

issue #1193


Steph4104:
I can't replicate this error on my machine. I will merge it and see on monday if I can replicate this error on prod.

Issue #966, opened 2017-04-12
Display search query box in Opportunity Platform


ToferC:
The search results page on the Opportunity Platform doesn't display the terms it is searching or give the user any way to modify or change their search other than clicking "back" or again selecting "search opportunity" and starting from scratch.

We should include the search bar with terms or at minimum display what terms are being searched.  We could also include a line clearly showing what filters are currently active as these are hidden in the drop downs and not immediately visible to users.

User Story: 

As a regular user, I need information on what I have searched for and options to refine or change the search so that I can better find the information that I'm looking for.
![searh](https://cloud.githubusercontent.com/assets/6226762/24958286/d12101b6-1f5c-11e7-8c33-016bec0c8f6f.PNG)

ToferC:
Would also be good to see about reducing real estate required for info box.

PR #1244, opened 2017-07-12
Filter opportunities platform

Steph4104:
Show which filter user use in the opportunities platform search

Issue: #966

### DIFF
--- a/mod/missions/pages/missions/display-search-set.php
+++ b/mod/missions/pages/missions/display-search-set.php
@@ -120,17 +120,17 @@ if ($search_form ) {
     );
 
 if ($advanced_form){
-    $content .='<strong>'.elgg_echo('missions:search_value').':</strong> ';
+    
+    $content .='<div class="mrgn-bttm-lg"><strong>'.elgg_echo('missions:search_value').':</strong> ';
 	for ($s = 0; $s < $number_of_rows; $s ++) {
         if ($advanced_form['selection_'.$s]){
 
-        	$value =str_replace('missions:', '', $advanced_form['selection_'.$s.'_element']);
-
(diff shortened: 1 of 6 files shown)


mod/missions/pages/missions/display-search-set.php
```diff
@@ -120,17 +120,17 @@
     );
 
 if ($advanced_form){
-    $content .='<strong>'.elgg_echo('missions:search_value').':</strong> ';
+    
+    $content .='<div class="mrgn-bttm-lg"><strong>'.elgg_echo('missions:search_value').':</strong> ';
```
piet0024: can we make this mrgn-bttm-sm? I feel like it takes up a lot of space and it will bring the "Clear" and "Refine" options closer

mod/missions/views/default/forms/missions/sort-missions-form.php
```diff
@@ -141,4 +142,39 @@
         </div>
 		
 	</div>
+    
+    <div class="col-sm-11 mrgn-tp-lg">
+        <strong><?php echo elgg_echo('missions:filter'); ?></strong>
```
piet0024: If there are no current filters can we hide this and only have it display when there are filters.

mod/missions/views/default/forms/missions/sort-missions-form.php
```diff
@@ -141,4 +142,42 @@
         </div>
 		
 	</div>
+    
+    <div class="col-sm-11 mrgn-tp-lg">
```
piet0024: Can this mrgn-tp-lg only be added when the filter string. Right now there is a large blank space when I am not filtering anything.

PR #1275, opened 2017-07-19
chore: use NodeJS 6 on Travis-ci

nschonni:
The recent build image updates seems to have cause issues with some of the old node_dependencies
https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming
/cc @pandurx

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: php
 php:
  - 5.6
 
+node_js:
+ - 6
+
 # This helps builds go quicker on Travis since it enables caching of dependencies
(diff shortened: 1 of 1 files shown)


Phanoix:
doesn't seem to help, I have it using dist: precise for now just for the node part until we get a better solution.

nschonni:
Worked on mine https://travis-ci.org/nschonni/gcconnex/builds/255391747

nschonni:
Looking at the builder info it looks like my form was still running precise too

PR #1284, opened 2017-07-21
chore: add EditorConfig for whitespace consistency

nschonni:
Opened an upstream issue for this but haven't heard back https://github.com/Elgg/Elgg/issues/10973

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+insert_final_newline = true
+indent_style = tab
+indent_size = 4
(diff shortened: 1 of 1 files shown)


nschonni:
@pandurx this might help some of your formatting cleanup

nschonni:
Many IDEs have direct support now and for those that don't there are plug-ins http://editorconfig.org/#download

nschonni:
> Is Travis also checking formatting?

Upstream has some stuff, but I don't believe it is setup for all the mods that have been added here.
This will also work with the GitHub web editor

pandurx:
been drowning in an endless sea of issues ...blah
I'm not sure how I would be able to test this, but this should support the editors that we are currently using (Sublime Text 2, Vim, and maybe nano)

PR #1289, opened 2017-07-24
minor fixes from the release for search patch

pandurx:


mod/gc_group_layout/views/default/groups/profile/summary.php
```diff
@@ -75,7 +75,7 @@
 
                 	} else {
 
-		                $group_name = ($group->title3) ? gc_explode_translation($group->title3, $lang) : $group->name;
+		                $group_name = ($group->name) ? gc_explode_translation($group->name, $lang) : $group->name;
```
Phanoix: the false condition is unreachable in this case and ends up behaving like gc_explode_translation() anyways.

mod/wet4/views/default/page/elements/head.php
```diff
@@ -66,13 +66,14 @@
     }
 
   }
+
 } else {
 
   $current_url = explode('/', $_SERVER['REQUEST_URI']);
   $page_entity = get_entity($current_url[count($current_url) - 1]);
 
   if ($elgg_entity instanceof ElggUser || $elgg_entity instanceof ElggGroup) {
```
nschonni: Aside: Just realized this conditional should be split since you're doing a ternary using the conditions inside.
```php
if ($elgg_entity instanceof ElggUser) {
  $page_title = $vars['title'];
} elseif ($elgg_entity instanceof ElggGroup) {
  $page_title = gc_explode_translation($elgg_entity->name, $gc_language) . " : {$vars['title']}";
} else {
...
```

pandurx: What am I splitting?

nschonni: The current `OR`'d condition `if ($elgg_entity instanceof ElggUser || $elgg_entity instanceof ElggGroup) {` since it contains a ternary with the same terms inside.
EX:
```
if (cond1 || cond2) {
  result = cond1 ? value1 : value2
}
...
```
Is the same as
```
if (cond1) {
  result = value1
}
else if (cond2) {
  result = value2
}
```
Unless i'm missreading it

pandurx: no, i mean, where in the code?

PR #1312, opened 2017-07-28
Twitter widget patch

pandurx:
Creating a new widget for new Twitter embed code

### DIFF
--- a/mod/widget_manager/views/default/widgets/twitter_widget/edit.php
+++ b/mod/widget_manager/views/default/widgets/twitter_widget/edit.php
@@ -26,24 +26,3 @@ $height = sanitise_int($widget->height, false);
 </div>
 
 
-<!--
-
-Collection
-<a class="twitter-timeline" href="https://twitter.com/TwitterDev/timelines/539487832448843776">National Park Tweets - Curated tweets by TwitterDev</a> <script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>
-
-Tweet
-Profile
-List
-Likes
-Handles
-Hashtags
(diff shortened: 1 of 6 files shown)


mod/widget_manager/views/default/widgets/twitter_widget/edit.php
```diff
@@ -0,0 +1,28 @@
...
+
+<?php // help link to create the twitter widget on Elgg ?>
+<div class='elgg-subtext'>
+	<?php echo elgg_view("output/url", array("href" => "https://twitter.com/settings/widgets", "target" => "_blank", "text" => elgg_echo("widgets:twitter_search:embed_code:help"))); ?>
+</div>
+
+<div>
+	<?php // Plain text box for the Twitter Widget embed code ?>
+	<p> <?php echo elgg_echo("widgets:twitter_search:embed_code"); ?> </p>
+	<p> <?php echo elgg_view("input/plaintext", array("name" => "params[embed_code]")); ?> </p>
```
Phanoix: The value for this and the type (dropdown below) should probably be loaded up when users edit the widget.

nschonni:
Might want to take a look at their docs for when there may be multiple instances on a page https://dev.twitter.com/web/javascript/loading

mod/widget_manager/views/default/widgets/twitter_widget/edit.php
pandurx: I might re-work this to save the code entirely, but the problem with that is a plugin that we've enabled (htmlawed) may strip that code.

PR #1359, opened 2017-08-21
cli install script for docker

Phanoix:
Installs a dev environment of GCconnex with all mods activated and ordered as in production.
creates an admin user with username/password: admin / adminpassword

to use:
once both web and db containers are running, run install with:
```
sudo docker exec -it <container name> php /var/www/html/install/cli/docker_installer.php
```
if you're not using this on localhost or want to use a port other than 8080 you will need to change the WWWROOT environment variable.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -61,6 +61,8 @@ services:
   # ###########################################################################
   gcconnex-db:
     image: library/mysql:5.6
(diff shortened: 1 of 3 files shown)


Phanoix:
yep, a collab flag is the next step now that this one's working

Issue #1331, opened 2017-08-07
Text snippet for discussions showing up as JSON when translation available


CalvinRodo:
Seems to be related to gctools-outilsgc/gcconnex#1313 
![json description](https://user-images.githubusercontent.com/87738/29042994-4d2bf24e-7b87-11e7-9172-8e7a737c11b3.png)

ToferC:
Thanks, @calvinrodo - approved for work. Please review the related cards and let's figure out why this issue is still occurring.

LemieuxGen:
@Steph4104 , will the fix also address the JSON that appears in the notifications? 
![json_notification digest_aug 31-2017](https://user-images.githubusercontent.com/15378977/29922581-35e23eea-8e24-11e7-8dc2-3ec52a95931c.JPG)

Steph4104:
@LemieuxGen  No, my fix is only for the view of the description of the discussion on the web site.

LemieuxGen:
Thanks @Steph4104 . I'll open a new ticket for the JSON in notifications.

CalvinRodo:
This is still occurring instead of JSON it's some other format: 
![bugpic1](https://user-images.githubusercontent.com/87738/30073595-570e0ec0-923c-11e7-8d0b-c103fabb005e.png)
![bugpic2](https://user-images.githubusercontent.com/87738/30073599-59cb2210-923c-11e7-928e-c1618a5fea24.png)

CalvinRodo:
It's also still appearing in the Latest Discussion list:
![bugpic3](https://user-images.githubusercontent.com/87738/30073621-78de5b86-923c-11e7-833a-09d1c7f7f413.png)

Steph4104:
The fix will be up on prod this week end

PR #1369, opened 2017-08-29
Fix description in discussion

Steph4104:
remove description3 from discussion

fix issue #1331


mod/wet4/views/default/object/groupforumtopic.php
```diff
@@ -25,14 +25,9 @@
 	}
 	return;
 }
+	$excerpt = elgg_get_excerpt(gc_explode_translation($topic->description, $lang));
```
nschonni: Indentation is off

PR #1372, opened 2017-08-29
group mail processing notification too slow

pandurx:


mod/cp_notifications/actions/cp_notifications/group_mail.php
```diff
@@ -0,0 +1,105 @@
...
+	
+	if (!empty($group) && ($group instanceof ElggGroup)) {
+		if ($group->canEdit()) {
+			set_time_limit(0);
+			
+			$body .= PHP_EOL . PHP_EOL;
+			$body .= elgg_echo("group_tools:mail:message:from") . ": {$group->name} [{$group->getURL()}]";
+			
+			// cyu - 03/07/2016: we need to use the cp_notifications module to handle the improved notifications
+			if (elgg_is_active_plugin('cp_notifications')) {
```
nschonni: Since this code is inside the `cp_notifications` module, can this actually ever be false?

mod/cp_notifications/actions/cp_notifications/group_mail.php
```diff
@@ -0,0 +1,105 @@
...
+
+
+/**
+ * helper function to retrieve all the members (and members who are selected to receive the group mail)
+ * @param int $group_id - group guid
+ * @param array $user_guids - array of user guids
+ */
+function group_tools_retrieve_members($group_id, $user_guids) {
+
+	$query = "SELECT guid_one FROM elggentity_relationships WHERE relationship = 'member' AND guid_two = '{$group_id}'";
```
nschonni: Not sure if Elgg handles SQL injection before 2.2 https://github.com/Elgg/Elgg/commit/a9e516826f8ee8bb763e12b3451aee96436ae985 but it might be good to check

mod/cp_notifications/actions/cp_notifications/group_mail.php
```diff
@@ -0,0 +1,105 @@
...
+	$group = get_entity($group_guid);
+	
+	if (!empty($group) && ($group instanceof ElggGroup)) {
+		if ($group->canEdit()) {
+			set_time_limit(0);
+			
+			$body .= PHP_EOL . PHP_EOL;
+			$body .= elgg_echo("group_tools:mail:message:from") . ": {$group->name} [{$group->getURL()}]";
+			
+			// cyu - 03/07/2016: we need to use the cp_notifications module to handle the improved notifications
```
nschonni: Date/comment might have been from original change

mod/cp_notifications/actions/cp_notifications/group_mail.php
pandurx: I basically just made an exact carbon copy of the original file, then slightly modified it

mod/cp_notifications/actions/cp_notifications/group_mail.php
pandurx: Elgg has a function that can handle queries. See ` get_data($query)` in the Elgg documentation

Phanoix: $group_id input gets cast as an int and not modified after that, so it should be fine.

PR #1373, opened 2017-08-29
removed single quotes around vote value field

### LINKED ISSUE #1355: Total Poll votes do not match distribution of votes in results
As a user, when I create a poll, the total number of votes should equal the combined results of the answers.

I can see the total number of votes are 83, but the results only show me 52 (40+12). Either some of the vote results are not displaying, or the total number of votes is skewed.
image

Link to Poll where this is occurring: https://gcconnex.gc.ca/polls/view/29629880/enwhat-would-you-call-an-online-inventory-of-former-students-eligible-for-integration-in-the-federal-public-service-of-canadafrquel-nom-donneriez-vous-u00e0-un-ru00e9pertoire-en-ligne-danciens-u00e9tudiants-salariu00e9s-admissibles-u00e0-lu2019intu00e9gration-dans-la-fonction-publique-fu00e9du00e9rale-du-canada

### LINKED ISSUE #1363: BUG - polling functionality doesn't accurately record all votes.
Voting for Option A in this [poll](https://gcconnex.gc.ca/polls/view/29629880/enwhat-would-you-call-an-online-inventory-of-former-students-eligible-for-integration-in-the-federal-public-service-of-canadafrquel-nom-donneriez-vous-u00e0-un-ru00e9pertoire-en-ligne-danciens-u00e9tudiants-salariu00e9s-admissibles-u00e0-lu2019intu00e9gration-dans-la-fonction-publique-fu00e9du00e9rale-du-canada) is not recording. After having selected and saved the vote (and refreshing the page),  the results show 0 votes for Option A. 

Other users have mentioned encountering the same issue.

Phanoix:
Single quotes were the problem with #1355 when the poll option had them as well.

Vote counts that were already affected by this can't be automatically recovered with certainty, but should be doable manually for most cases if necessary.

fixes #1355 and fixes #1363

### DIFF
--- a/mod/wet4/views/default/forms/polls/vote.php
+++ b/mod/wet4/views/default/forms/polls/vote.php
@@ -50,7 +50,7 @@ foreach ($responses3 as $value) {
 	$responses4 = gc_explode_translation($value, $lang);
 
 
-$form_body .="  <input type='radio'  class = 'mrgn-rght-sm' id='responses-".$responses4."' name='response3' value='".$value."'> <label for='responses-".$responses4."'>".$responses4."</label><br>";
+$form_body .="  <input type='radio'  class = 'mrgn-rght-sm' id='responses-".$responses4."' name='response3' value=".$value."> <label for='responses-".$responses4."'>".$responses4."</label><br>";
 }
 
 
--- a/mod/wet4/views/default/forms/polls/vote.php
+++ b/mod/wet4/views/default/forms/polls/vote.php
@@ -50,7 +50,7 @@ foreach ($responses3 as $value) {
 	$responses4 = gc_explode_translation($value, $lang);
 
 
-$form_body .="  <input type='radio'  class = 'mrgn-rght-sm' id='responses-".$responses4."' name='response3' value=".$value."> <label for='responses-".$responses4."'>".$responses4."</label><br>";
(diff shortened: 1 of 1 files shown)


mod/wet4/views/default/forms/polls/vote.php
```diff
@@ -50,7 +50,7 @@
 	$responses4 = gc_explode_translation($value, $lang);
 
 
-$form_body .="  <input type='radio'  class = 'mrgn-rght-sm' id='responses-".$responses4."' name='response3' value='".$value."'> <label for='responses-".$responses4."'>".$responses4."</label><br>";
+$form_body .="  <input type='radio'  class = 'mrgn-rght-sm' id='responses-".$responses4."' name='response3' value=".$value."> <label for='responses-".$responses4."'>".$responses4."</label><br>";
```
nschonni: `value='". htmlEntities($value, ENT_QUOTES) ."'` seems to be the safe way to do this

Phanoix: The polls mod for whatever reason does string matching to figure out what option a vote is for... so the value would need to be identical here to the json string that's used to store the text for the vote response/option. I checked some of the more recent versions and they all seem to stick to this way of storing votes.

nschonni: Wouldn't any value with a space blow up then? Or is it better to filter out poll options with any single quotes instead.

Phanoix: Did a bit of testing and it looks like elgg does actually parse the html entities in the value before storing it so your suggestion would the the way to go, thanks!

PR #1374, opened 2017-08-30
clean up search results

pandurx:


mod/b_extended_profile/views/default/profile/wrapper.php
```diff
@@ -81,26 +85,27 @@
...
         }
 
-        //if(elgg_is_logged_in()){
-            if ( has_content($user, 'gc_skills') ) {
-                init_ajax_block(elgg_echo('gcconnex_profile:gc_skills'), 'skills', $user);
-                echo elgg_view('b_extended_profile/skills');
-                finit_ajax_block('skills');
-            }
-        //}
+        if ( has_content($user, 'gc_skills') && strstr(strtolower($_SERVER['HTTP_USER_AGENT']), 'gsa-crawler') === false ) {
```
nschonni: There is apparently [`stristr`](http://php.net/manual/en/function.stristr.php) for the case insensitive version.

mod/wet4/views/default/page/elements/topbar_wrapper.php
```diff
@@ -23,9 +23,7 @@
 
 if (!elgg_is_logged_in()) {
     echo '<nav role="navigation" class="text-right col-md-12 mrgn-tp-sm user-z-index">';
-
```
nschonni: Might be good to rebase these changes out since it is just whitespace stuff

mod/b_extended_profile/views/default/profile/wrapper.php
nschonni: You fixed all of them, but the one I commented on :stuck_out_tongue:

Issue #1307, opened 2017-07-26
H1 on bookmark page missing owner's name


nschonni:
1. Going to a bookmark page for a page. EX: https://gcconnex.gc.ca/bookmarks/owner/Chris_Allison
2. Note that the H1 contains only `'s bookmarks`, where it should likely be `Username's bookmarks`

![image](https://user-images.githubusercontent.com/1297909/28646664-e60957dc-7230-11e7-9fbc-86d72b435baa.png)

```html
<h1 property="name">'s bookmarks</h1>
```

ToferC:
This might extend to other content types.

PR #1384, opened 2017-09-05
H1 in owner bookmarks view

Steph4104:
Change title to name to get the name of the owner.
Fix issue #1307

### DIFF
--- a/mod/bookmarks/pages/bookmarks/owner.php
+++ b/mod/bookmarks/pages/bookmarks/owner.php
@@ -10,8 +10,8 @@ if (!$page_owner) {
 	forward('', '404');
 }
 
-$title_bookmarks = elgg_echo('bookmarks:owner', array(gc_explode_translation($page_owner->title, $lang)));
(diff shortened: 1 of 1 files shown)


mod/bookmarks/pages/bookmarks/owner.php
```diff
@@ -10,8 +10,8 @@
 	forward('', '404');
 }
 
-$title_bookmarks = elgg_echo('bookmarks:owner', array(gc_explode_translation($page_owner->title, $lang)));
-$title = elgg_echo(gc_explode_translation($page_owner->title, $lang));
+$title_bookmarks = elgg_echo('bookmarks:owner', array(gc_explode_translation($page_owner->name, $lang)));
+$title = elgg_echo(gc_explode_translation($page_owner->name, $lang));
```
nschonni: Does the `gc_explode_translation` make sense now that this is talking about a name?

Steph4104: Name can be in both language

Issue #1241, opened 2017-07-11
Create Group auto-suggest of existing groups not working


ToferC:
We previously developed functionality so that when a user is creating a group, we would provide a list of similarly named groups and suggest that they may want to join an existing group rather than create a new one.

This functionality is no longer working and should be fixed as part of our summer upgrade.

User Story: As a regular users, I need to know if other groups are already meeting the need that I see so that I and other can contribute expertise and content in the most effective way possible.

As an administrator or regular user, I benefit from having content and discussions in themed areas rather than spread across the platform.

LemieuxGen:
@Steph4104 Nous avons jeter un coup d'oeil ce matin à GCconnex, et l'option "Auto suggest" ne fonctionne pas. Est-ce que ça l'a été relâché sur GCconnex cette fin de semaine? Merci.

PR #1388, opened 2017-09-06
Add autocomplete for group name

Steph4104:
When a user create a group, they have the autocomplete function for the name. It search in both language and show you the list of group name that have in the language of your page.

Should fix issue #1241

### DIFF
--- a/mod/wet4/views/default/input/groups_autocomplete.php
+++ b/mod/wet4/views/default/input/groups_autocomplete.php
@@ -19,19 +19,24 @@
 	elgg_load_css("group_tools.autocomplete");
 	
 	$site_url = elgg_get_site_url();
-?>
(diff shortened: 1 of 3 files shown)


mod/wet4/views/default/input/groups_autocomplete.php
```diff
@@ -52,19 +57,19 @@
 					source: function( request, response ) {
 						$.getJSON( "<?php echo $site_url; ?>groups_autocomplete", {
 							q: request.term,
-							'groups_guids': function() {
-								var result = "";
+							 'groups_guids': function() {
```
nschonni: Looks like the rest of the changes below are an accidental whitespace change that added a space in front of the tabs

PR #1418, opened 2017-09-19
If user is delete, the group is create is not delete if it change the owner

Steph4104:
change container guid and metadata when owner change

### DIFF
--- a/mod/group_operators/actions/group_operators/change_owner.php
+++ b/mod/group_operators/actions/group_operators/change_owner.php
@@ -44,9 +44,7 @@
 
             if ($metadata){
 
-                foreach ($metadata as $data){
-					update_data("UPDATE {$db_prefix}metadata SET owner_guid = '$who_guid' where id = $data->id");
-                }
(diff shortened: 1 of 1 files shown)


mod/group_operators/actions/group_operators/change_owner.php
```diff
@@ -36,6 +37,18 @@
...
 		$mygroup->owner_guid = $who_guid;
+        $mygroup->container_guid = $who_guid;
+
+        //Update metadata owner guid is case user creator is delete
+ 		$metadata = get_data("SELECT * FROM {$db_prefix}metadata WHERE entity_guid =$mygroup_guid");
+
+            if ($metadata){
+
+                foreach ($metadata as $data){
+					update_data("UPDATE {$db_prefix}metadata SET owner_guid = '$who_guid' where id = $data->id");
```
Phanoix: Should be able to do without the loop with:
```sql
UPDATE {$db_prefix}metadata SET owner_guid = '$who_guid' where entity_guid = $mygroup_guid
```

PR #1419, opened 2017-09-19
Fixing edit with source editor

### LINKED ISSUE #1086: Changes not saving when editing content using "Source" editing and when pasting content in Rich text editor
As a user I need to be able to save Edits to content using "Source" mode and when pasting new content using the Rich text editor.
When I edit a page or discussion using the "Source" mode, the changes will not save.
If the change to the content is typed or is a copy/paste from a webpage, it will save in the Rich text editor however if it is copy/paste from Word or an e-mail, it will not save.

ethanWallace:
Fixes #1086


markwooff:
That's all it was?! Awesome, thanks for fixing @ethanWallace. Been a frustrating issue.

ethanWallace:
@markwooff I found something where it wasn't working if the user did something with the english/french entry. Could you test it again?

markwooff:
Updated code works with both French & English source content updates. Might be a good idea to test on GCconnex as well.

Issue #1326, opened 2017-08-03
Delete icon visible for users not logged in on Wire posts


gabriellehubert:
When a user is browsing the Wire while they are not logged in, they will see the delete (or trash can) icon visible below each Wire post (see first image). Although clicking this icon does not allow the user to delete the post (see second image), it can be confusing for a user to see an icon that gives this functionality.

![wire post bug](https://user-images.githubusercontent.com/25649583/28922498-6112e8d8-7828-11e7-9138-a89e5def3f49.PNG)

![wire post bug 2](https://user-images.githubusercontent.com/25649583/28922413-11c211d2-7828-11e7-8a0c-47da9748d8a3.PNG)

PR #1420, opened 2017-09-19
Hide delete icon if not logged in

Steph4104:
if logged in show icon and delete link.

Issue #1326

### DIFF
--- a/engine/lib/navigation.php
+++ b/engine/lib/navigation.php
@@ -483,16 +483,18 @@ function _elgg_entity_menu_setup($hook, $type, $return, $params) {
 		);
 		$return[] = \ElggMenuItem::factory($options);
 
-		// delete link
-		$options = array(
-			'name' => 'delete',
-			'text' => elgg_view_icon('delete'),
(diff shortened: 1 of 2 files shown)


engine/lib/navigation.php
```diff
@@ -483,16 +483,18 @@ function _elgg_entity_menu_setup($hook, $type, $return, $params) {
...
-		$options = array(
-			'name' => 'delete',
-			'text' => elgg_view_icon('delete'),
-			'title' => elgg_echo('delete:this'),
-			'href' => "action/$handler/delete?guid={$entity->getGUID()}",
-			'confirm' => elgg_echo('deleteconfirm'),
-			'priority' => 300,
-		);
-		$return[] = \ElggMenuItem::factory($options);
+		if(elgg_is_logged_in()){
```
nschonni: Maybe extend the same check from L464 to cover the rest of the function body instead

Steph4104:
If I only modify the wet4, It show the delete bottom from the core.

PR #1423, opened 2017-09-19
tags will link to a GSA query

### LINKED ISSUE #537: Hyperlink tags to search

piet0024:
When a user clicks on the tags it will now query the GSA 

Closes #537

### DIFF
--- a/mod/gcconnex_theme/views/default/output/tags.php
+++ b/mod/gcconnex_theme/views/default/output/tags.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * Elgg tags
+ * Tags can be a single string (for one tag) or an array of strings. Accepts all output/tag options
+ *
+ * @uses $vars['value']      Array of tags or a string
+ * @uses $vars['entity']     Optional. Entity whose tags are being displayed (metadata ->tags)
+ * @uses $vars['list_class'] Optional. Additional classes to be passed to <ul> element
+ * @uses $vars['item_class'] Optional. Additional classes to be passed to <li> elements
+ * @uses $vars['icon_class'] Optional. Additional classes to be passed to tags icon image
+ */
+ /*
+ * GC_MODIFICATION
+ * Description: Added font awesome icon for start of tags list. Custom styling to tags
+ * Author: GCTools Team
+ */
+if (isset($vars['entity'])) {
+	$vars['tags'] = $vars['entity']->tags;
+	unset($vars['entity']);
+}
+
+$value = elgg_extract('value', $vars);
+unset($vars['value']);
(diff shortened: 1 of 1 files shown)


mod/gcconnex_theme/views/default/output/tags.php
```diff
@@ -0,0 +1,84 @@
...
+	}
+}
+
+if (empty($list_items)) {
+	return;
+}
+
+$icon = "<span class=\"fa fa-tags fa-lg\"></span>";
+
+$list = <<<___HTML
```
pandurx: you can probably just echo this portion of the HTML code

mod/gcconnex_theme/views/default/output/tags.php
```diff
@@ -0,0 +1,84 @@
...
+	$item_class = "$item_class {$vars['item_class']}";
+	unset($vars['item_class']);
+}
+
+$icon_class = elgg_extract('icon_class', $vars);
+unset($vars['icon_class']);
+
+$list_items = '';
+
+$params = $vars;
```
pandurx: the variable `$params` is not used anywhere, you should take this out

mod/gcconnex_theme/views/default/output/tags.php
```diff
@@ -0,0 +1,84 @@
...
+        //Adding GSA query to tags
+        $query = '?q='.$tag.'&a=s&s=3&chk4=on';
+        $url = "http://intranet.canada.ca/search-recherche/query-recherche-" .get_current_language(). ".aspx" .$query;
+		$list_items .= "<li class=\"$item_class\">";
+		$list_items .= '<a href="'.$url.'">'.$tag.'</a>';
+		$list_items .= '</li>';
+	}
+}
+
+if (empty($list_items)) {
```
pandurx: i think you already do this check at the beginning of the file, check if `$tags` is empty (or is not an array)

mod/gcconnex_theme/views/default/output/tags.php
```diff
@@ -0,0 +1,82 @@
...
+if (empty($vars['tags']) && (!empty($value) || $value === 0 || $value === '0')) {
+	$vars['tags'] = $value;
+}
+
+if (empty($vars['tags']) && $value !== 0 && $value !== '0') {
+	return;
+}
+
+$tags = $vars['tags'];
+unset($vars['tags']);
```
pandurx: `$tags = $vars['tags'];
unset($vars['tags'];`

PR #1426, opened 2017-09-20
Group autocomplete prod

Steph4104:
Fix group autocomplete

### DIFF
--- a/mod/wet4/lib/groups_autocomplete.php
+++ b/mod/wet4/lib/groups_autocomplete.php
@@ -28,7 +28,6 @@
 		$join = "JOIN {$db_prefix}groups_entity ge ON e.guid = ge.guid";
 		$params['joins'] = array($join);
 		$fields = array('name', 'description');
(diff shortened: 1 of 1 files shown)


mod/wet4/lib/groups_autocomplete.php
```diff
@@ -26,7 +26,7 @@
 		$join = "JOIN {$db_prefix}groups_entity ge ON e.guid = ge.guid";
 		$params['joins'] = array($join);
 		$fields = array('name', 'description');
-		$where = search_get_where_sql('ge', $fields, $params,FALSE);
+		$where = "ge.name LIKE '%$q%'";
```
Phanoix: This only searches name as opposed to name and description like, and the search function would have used $params['query']. Should do more or less what it's supposed to anyways though.

Steph4104: I don't think we need to search into the description.

Issue #1422, opened 2017-09-19
"Related Groups" still shows on right hand side of group page even if option disabled


KevinChagnon:
"Related Groups" still shows on right hand side of group page even if option disabled under Content Options;

https://gcconnex.gc.ca/groups/profile/24813917/encpab-central-centrale-dgcapfrcentrale-dgcap

![related groups](https://user-images.githubusercontent.com/23527358/30607157-0cf148dc-9d42-11e7-8219-8ae53c25d4a7.jpg)

![related groups settings](https://user-images.githubusercontent.com/23527358/30607176-1b575a56-9d42-11e7-910a-662f46989822.PNG)

KevinChagnon:
The "Show Related Groups" option under Content Options should be renamed to "Show Related Groups in side bar" and should not add/remove the option of seeing a link to "Related Groups" under "More" on the group page.

ToferC:
Minor issue. Please do not action until redesign.

KevinChagnon:
https://github.com/Phanoix/gcconnex-help/issues/27

PR #1430, opened 2017-09-21
Related group option hide in the sidebar

Steph4104:
The option of related group show or hide the section in the sidebar instead of the link under "more"

Issue #1422


Steph4104:
The widget of related group work without the checkbox. 
If the checkobox is unchek but a user add a related group with the widget, the related group in the sidebar will appears. 

The group option can appear under the More dropdown, but only operator or owner can add related group.

PR #1434, opened 2017-09-21
Language issue in close group

Steph4104:
Change 'se' to 'ce' because it's demonstrative.

Issue #1400


pandurx:
perhaps, for a really minor language change such as this, we don't need someone to review

PR #1435, opened 2017-09-21
Merging all GCcollab code into GCconnex repo

markwooff:
Includes all GCcollab modules & updates including:
- Additional modules used by GCcollab (`gc_mobile_api`, `loginrequired`, `elgg_solr`, latest version of API Admin, etc.)
- GCcollab-specific override modules as to not edit existing code


nschonni:
For those 3rd party plug-ins, wouldn't it make more sense to use composer instead of adding all this into the base?

markwooff:
Yeah, I think the main culprit is the `linkedin_profile_importer` module. @nschonni agreed that we should be using composer, we're working on getting there...

PR #1450, opened 2017-09-26
Freshdesk type input

### LINKED ISSUE #1446: Add fields for freshdesk tickets=
As a help desk agent, I need users to be able to identify the type of request when they are submitting a ticket so that I can better answer their request and avoid too many back and forth with the user. 

Add new field in ticket widget (Type).

ethanWallace:
Adds type input to ticket submit form to help help desk identify issues easier.

Closes #1446

### DIFF
--- a/mod/contactform/index.php
+++ b/mod/contactform/index.php
@@ -1,7 +1,9 @@
 <?php
 // Load Elgg engine
 include_once(dirname(dirname(dirname(__FILE__))) . "/engine/start.php");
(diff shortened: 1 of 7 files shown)


ethanWallace:
I agree that being able to be add a more specific type while other is selected would be ideal. The choices in the type input are based on the ticket filter options on the Freshdesk site so it wouldn't be possible to add custom types while other is selected.

PR #1451, opened 2017-09-26
Adding codecov upload to tavis ci and a code coverage badge

Phanoix:
We don't have a scrutinizer subscription, replacing it with codecov


nschonni:
Just and FYI, https://scrutinizer-ci.com/pricing

> What about open-source repositories?

> We do not charge for open-source repositories, you can add as many as you like.

Phanoix:
Wow, they really berried that down there... Since codecov doesn't seem to want to work I'll try scrutinizer since elgg is already using it.

Issue #1431, opened 2017-09-21
Opportunities Platform > Find Members > Advance Find


LemieuxGen:
1. When ‘Field’ “Skill” is selected; the Value box does not offer the auto-suggest from the skills list anymore. [Connex & Collab]
2. When ‘Field’ “User department” is selected; the Value box does not offer the drop down list of departments used in profiles for consistent spelling. [Connex & Collab]
3. When ‘Field’ “Opt In” is selected; the Value box does not offer the drop down list of mechanisms offered via the Opportunities Platform anymore – it is a free text box. [Connex; resolved on Collab]
4. Searches still max out at 100 entries, but we need it to show the actual total number for depts using this feature to report how many employees in their departments are use it for various mechanisms. Also will be required for our own reporting, including for the Clerk’s Report.

mbujold:
@LemieuxGen 
Tested issues:
- [ 1 ] GCConnex & GCCollab both showed auto-suggested options for skills - Can't reproduce.
- [ 4 ] Can you elaborate on this a bit more so I can truly test this?

marcleblanc:
@LemieuxGen Please let us know feedback on what @mbujold mentioned about issues 1 and 4. Thanks!

LemieuxGen:
Received an response back from OCHRO: 
**•[ 1 ] GCConnex & GCCollab both showed auto-suggested options for skills - Can't reproduce.**
This issue is ongoing on both GCcollab & GCconnex.  See the screen shot below. There should be an autosuggest list popping up with skills that begin with the letter “L”.  I have tried from a few computers and the issue is consistent.  Make sure when trying to reproduce that you go to Opportunities Platform > Find Members > Advance Find
![op advanced search 1](https://user-images.githubusercontent.com/15378977/31079144-9be27f24-a752-11e7-94b0-47072a420f45.png)

**•[ 4 ] Can you elaborate on this a bit more so I can truly test this?**
 Search for a term like “communication” or “project” and you will see a notification that the system has reached the maximum number of entries (which is 100). There should not be a maximum number, we need to know the actual number of users who have identified certain skills, or opted in for certain mechanisms. This is required for our reporting responsibilities.  In addition to showing the all the relevant profiles, it would be very helpful to have a count at the top of the page of how many entries were found (like on amazon, example: “there are 1,957 entries that match your search criteria”.

marcleblanc:
[...]

PR #1464, opened 2017-09-28
Opportunities advanced find members 1431

marcleblanc:
Updated the advanced form to have a drop down list of 'departments' when selecting the 'User department' field and also updated the 'Opt In' field to show a drop down list as well (used same options as Collab).

Updated the language files to have the proper key:value in them.

These changes are for fixing for the following bugs:

- When ‘Field’ “User department” is selected; the Value box does not offer the drop down list of departments used in profiles for consistent spelling. [Connex & Collab]
- When ‘Field’ “Opt In” is selected; the Value box does not offer the drop down list of mechanisms offered via the Opportunities Platform anymore – it is a free text box. [Connex; resolved on Collab]

[Opportunities Platform > Find Members > Advance Find #1431](https://github.com/gctools-outilsgc/gcconnex/issues/1431)


mod/missions/views/default/missions/element-select.php
```diff
@@ -116,6 +116,22 @@
             ));
         	break;
             
+        case elgg_echo('missions:opt_in'):
+        	$content .= elgg_view('input/dropdown', array(
+                'name' => $dropdown_name . '_element',
+                'value' => '',
+                'options' => array(
```
markwooff: These are the opt-in options available on GCcollab, there are more available for GCconnex that are missing.

Phanoix:
travis is erroring out with : 
```
ReferenceError: Can't find variable: jQuery
at /home/travis/build/gctools-outilsgc/gcconnex/vendors/jquery/jquery-migrate-1.2.1.min.js:2
```

lucbelliveau:
@Phanoix I notice jquery was updated to version 1.11.0, possibly during the installation process automatically?  @marcleblanc ?  Or was it intentional to fix something?

Either way, I notice in the CI this:

> WARN [watcher]: Pattern "/home/travis/build/gctools-outilsgc/gcconnex/vendors/jquery/jquery-1.11.0.min.js" does not match any file.

Not sure where this came from, but it should probably be updated to 1.11.1.min.js.

Phanoix:
Seems to be hard-coded in:
* engine/lib/views.php
* js/tests/karma.conf.js
* views/installation/page/default.php

also came up in  vendors/jquery/jquery.min.map

marcleblanc:
@lucbelliveau The jQuery version update to 1.11.1 was to fix the IE issue with the drop downs as per markwoof's note in the issue:

> Note: just tested Item 1 on GCcollab, can confirm the auto-suggest appears on Chrome - however it does not on IE. This is the bug that will need to be fixed.

PR #1471, opened 2017-09-29
Optimized api

lucbelliveau:
@Phanoix if we can get this into prod ASAP... this will greatly improve the speed of the api... and we need to get all the data as soon as possible...


Phanoix:
I'll review this tomorrow and probably have it up by / on Monday.

lucbelliveau:
@Phanoix any idea when it will available on prod?

Phanoix:
it's up

PR #1472, opened 2017-10-02
Added new "Collaboration" opportunity type

markwooff:
New opportunity type only affecting GCcollab

### DIFF
--- a/mod/missions_collab/lib/missions-searching.php
+++ b/mod/missions_collab/lib/missions-searching.php
@@ -138,22 +138,22 @@ function mm_analyze_advanced_search_element($place, $array)
                     case elgg_echo('gcconnex_profile:opt:micro_mission'):
                         $name_option = 'opt_in_missionCreate';
                         break;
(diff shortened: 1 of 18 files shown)


mod/missions_collab/lib/missions-searching.php
```diff
@@ -200,6 +200,12 @@ function mm_analyze_advanced_search_element($place, $array)
                     case elgg_echo('gcconnex_profile:opt:student_create');
                         $name_option = 'opt_in_student_create';
                         break;
+                    case elgg_echo('gcconnex_profile:opt:collaboration_seek');
+                        $name_option = 'opt_in_collaboration_seek';
+                        break;
+                    case elgg_echo('gcconnex_profile:opt:collaboration_create');
```
Phanoix: case statements should end with ':' as Scrutinizer pointed out, but I guess it works anyways.

markwooff: I'll fix these up as this incorrect syntax exists in the `missions` mod as well.

PR #1478, opened 2017-10-02
1092 analytics

### LINKED ISSUE #1092: Opportunity Platform Analytics - "Offered" transaction doesn't appear
When a  posted opportunity is offered to an applicant (i.e. after the opportunity has been offered but before the applicant accepts or declines), the transaction type "offered" is not shown in the stacked graph separated by transactions. (tested in pre-prod)

lucbelliveau:
Fixes #1092 

This should resolve this issue... looks like the analytic queries were failing due to permissions ... when this is added to production, the history since Jun 21st should be available.


lucbelliveau:
The whitespaces are left-over from the original implementation... I usually don't fix them to produce cleaner diffs.. Since you've already reviewed the changes I'll clean them up for this file and push it up.

nschonni:
@lucbelliveau I meant that you've introduced whitespace issues with your changes that have made the diff larger

PR #1483, opened 2017-10-03
chore(au_subgroups): remove unessary breaks

nschonni:
Since there is a return statement before, the break is never hit


nschonni:
Oops, didn't realize this was a separate project, I thought it was anther GC custom one. Is this the correct upstream project https://github.com/AU-Landing-Project/au_subgroups

PR #1492, opened 2017-10-04
Forum access (display only)

pandurx:


mod/gcforums/views/default/forms/gcforums/create.php
```diff
@@ -80,7 +80,7 @@ function general_information_form($object = null) {
 		));
 		$sub_return = array('title' => array($lblTitle, $txtTitle));
 
-		$access_id = 2;
+		$access_id = (!$object->access_id) ? ACCESS_DEFAULT : $object->access_id;
```
nschonni: Not sure if the non-negated version would be easier to read

mod/gcforums/actions/gcforums/edit.php
```diff
@@ -58,6 +58,7 @@
 		// TODO: form for sticky
 		$entity->sticky = 0;
 		$entity->save();
+
```
nschonni: Might want to rebase out this file change

PR #1503, opened 2017-10-06
tool to allow topic moving to another forum

pandurx:


mod/gcforums/actions/gcforums/move_topic.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+$topics = str_replace(' ', '', get_input('txtTopic'));
+$forum = (int)get_input('txtForum');
+
+$error = "";
+
+// validate that the inputs are correct
+if (!preg_match("/[\d*,\s?]+/", $topics)) {
+	$error .= "<p>invalid input for topics</p>";
```
nschonni: If this is admin only it might not matter as much, otherwise maybe a translation string is needed. Few others below

pandurx: this is only for site administrator (so you and users of the site, will not have access) it has to do with the job marketplace

nschonni: Sounds reasonable

mod/gcforums/views/default/forms/gcforums/move_topic.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+
+$notice = "<p><form action='{$base_url}gcforums/move_topic'><div class='administrative-topic-tool'>
```
nschonni: nit: not sure the `<p>` should wrap the form

mod/gcforums/actions/gcforums/move_topic.php
```diff
@@ -0,0 +1,30 @@
...
+
+$topic_guids = explode(',', $topics);
+
+foreach ($topic_guids as $topic_guid) {
+	$topic_entity = get_entity($topic_guid);
+	$topic_entity->container_guid = $forum;
+	$topic_entity->save();
+}
+
+system_message("topic(s) were successfully moved!");
```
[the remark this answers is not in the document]
pandurx: see comment thread by @nschonni

mod/gcforums/views/default/forms/gcforums/move_topic.php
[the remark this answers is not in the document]
pandurx: i don't think it requires the security token as i'm calling it using the elgg native function that calls the form, it takes care of it

pandurx: do you know what the best uses for `<p></p>`? i've always thought `<br/>` was not the best thing to use for newlines

nschonni: I was looking at https://getbootstrap.com/docs/3.3/css/#forms, but inspecting the page I saw the the default for `<form>` is `display:block`. You could wrap it in a div, but I don't think it's necessary.

PR #1517, opened 2017-10-11
chore(missions): php-cs-fixer formatting

nschonni:
Manual removal of dead code and line trimming too.

This may replace #1510, but that one can land easier

Non-whitespace diff can be seen here https://github.com/gctools-outilsgc/gcconnex/pull/1517/files?w=1


Phanoix:
I think at this point it would be easier to rerun the cleanup rather than go through all the conflicts.

nschonni:
No problem, but I'm not going to resubmit them

PR #1529, opened 2017-10-13
Testing - Suspend user mod

### LINKED ISSUE #233: Add retired / suspended user functionality
Create better understanding of which users are active on the tools

piet0024:
Plugin that will allow users to delete / deactivate their account. Used https://elgg.org/plugins/860132 as a base mod.

Closes #233 

This mod is ready for testing, usability testing. Need language. Should be up on preprod soon.

User can go to account settings > Delete account and fill out the form. If the user is an owner of a group they must transfer that ownership before they can deactivate.

To reactivate a user the site admin will go to Admin Dashboard > Users > GC Deactivated Users to see and reactivate users.


mod/member_selfdelete/manifest.xml
```diff
@@ -0,0 +1 @@
...
    <version>2.0</version>
    <description>Allows members to delete their account, and submit an optional reason for deleting their account</description>
    <website>http://mattbeckett.me/</website>
    <copyright>See COPYRIGHT.txt</copyright>
    <license>GNU General Public License version 2</license>
    <requires>
        <type>elgg_release</type>
        <version>1.9</version>
    </requires>
</plugin_manifest>
```
Phanoix: Seems to have odd line endings here, though if it doesn't prevent the mod from activating it shouldn't really matter.

mod/member_selfdelete/views/default/admin/users/member_selfdelete/deactivated_users.php
```diff
@@ -0,0 +1,28 @@
...
+echo '<table class="elgg-table">';
+echo '<tr><th>Username</th><th>Email</th><th>Reason</th><th>State</th><th>Toggle</th></tr>';
+foreach ($users as $user){
+    echo '<tr>';
+    echo '<td>' .$user->username . '</td>';
+    echo '<td>' .$user->email . '</td>';
+    if($user->gcdeactivatereason){
+        echo '<td>' .elgg_echo('member_selfdelete:gc:reason:' . $user->gcdeactivatereason). '</td>';
+    }else{
+        echo '<td>No reason given</td>';
```
Phanoix: I suppose this should be fine since it's an admin only view.

PR #1534, opened 2017-10-15
Cleanup invitefriends

nschonni:
- php-cs-fixer autocleanup
- remove flagged variable issue from Codacy


nschonni:
Realized that this is a core plug-in, will close and see if anything should be upstreamed

PR #1535, opened 2017-10-16
remove/leave group removes the forum subscriptions

pandurx:


pandurx:
@nschonni i might've done it this way because i couldn't figure out how to fire the event hook after the action has successfully removed you from the group, and having to patch an issue in a short period of time.

in any case, it does make sense to implement a hook instead of modifying the plugin action file.

PR #1566, opened 2017-10-24
Fix typos

maazsq:


nschonni:
Might be better to fix the plural stuff at the top of the file friend->user
```php
if(empty($singular)){ $singular = elgg_echo('friend'); }
if(empty($plural)){ $plural = elgg_echo('friends'); }
```

PR #1589, opened 2017-11-01
forum url links in setting page

pandurx:


pandurx:
@Steph4104 
it would be an enhancement and we're not doing any more enhancement to the site. the opportunities platform notification doesn't follow with the rest of the content either because you don't "subscribe", users would email the person who offers/seeks posting anyways and you can't post comments in response to the posting

PR #1596, opened 2017-11-09
Closed group

pandurx:


Steph4104:
When user send a invitation with the colleague option, the text is only in english.
image

The french text is miss in fr.php

pandurx:
@Steph4104  @markwooff 
This patch should work on both GCcollab and GCconnex without any modification, will grab links and site title instead of static text

Steph4104:
For some reason, when I test on preprod, I'm not able to receive email when I invite a colleague to my close group.

Also, I test invite by email and on the french section, some variable are at the wrong place.

join_close_group

PR #1597, opened 2017-11-09
Adding images to the wire

markwooff:
New feature for adding images to Elgg wire posts


markwooff:
@ethanWallace the error on the FR language has been fixed. However you're right that some work into the accessibility side of things will need to be done.

Issue #1602, opened 2017-11-15
IE not playing embedded Flash videos from YouTube


klecuyer:
Linked to https://github.com/Phanoix/gcconnex-help/issues/31

As a user, I cannot view embedded YouTube videos in IE because GCconnex converts the code to Flash, which isn't supported in IE. I don't have this problem in Firefox or Chrome because those browsers automatically convert these files to playable MV5 format.

User's IT Security's response (ISED) - this is how the response appeared from the user's tech support:
![image](https://user-images.githubusercontent.com/29754185/32850589-95710d24-ca00-11e7-87a3-09902d83ae41.png)

Sebastien's note:
Embeding Youtube videos currently uses Flash, and this is causing problems for some users, depending on what browsers they have access to and if flash is supported on those browsers in departments.
Fix would probably be here (line 109 and a few others)
https://github.com/gctools-outilsgc/gcconnex/blob/gcconnex/mod/embed_extender/lib/embedvideo.php
Instead of using <object> we could use an <iframe> to embed Youtube videos without using Flash.
Si vous avez des questions, n’hésitez pas à communiquer avec moi.
Sébastien Lemay

klecuyer:
Initial email forwarded to @Troy-Lawson

PR #1603, opened 2017-11-15
Change object to iframe for Youtube videos

smellems:
Possible fix for #1602


KevinChagnon:
In our help material, when advising users which platform they can use to upload videos to then link to them in GCconnex, in addition to Youtube we also mention vimeo and Dailymotion

Could we look into fixing this so that links on GCconnex that point to these platforms  also work in IE?

@Phanoix @LemieuxGen

smellems:
This should now work for Youtube, Vimeo, Dailymotion and Metacafe.

Pull request submitted upstream: https://github.com/beck24/embed_extender/pull/12

klecuyer:
@piet0024 does this mean I can close #1602 or not yet?

PR #1632, opened 2017-11-28
use cp_notification to send the welcome message

### LINKED ISSUE #1502: Group welcome message not sending
Users are reporting that the group welcome message in GCconnex does not send.

I suspect this may be linked to notifications/emails.

Group welcome message is not working (does not send a message to new group members).
Attached is a screenshot of my group welcome message (found by clicking on settings - edit group - under the admin options tab).
image

Linked to https://github.com/Phanoix/gcconnex-help/issues/16

Steph4104:
Add the welcome message to the cp_notification function

Fix #1502

### DIFF
--- a/mod/cp_notifications/languages/en.php
+++ b/mod/cp_notifications/languages/en.php
@@ -320,6 +320,11 @@ $english = array(
 		%s <br/>
 		You can view or reply to this by clicking on this link: %s",
 
+	// welcome message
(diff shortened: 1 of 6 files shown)


ethanWallace:
Email seems to send out fine now when a user joins the group. One thing to perhaps look at as well is if the group owner uses [group_name] in the message the name will appear with the en/fr json around it.

PR #1645, opened 2017-12-06
First attempt at gcconnex kubernetes deployment file

Phanoix:
Initial version of gcconnex k8s deployment file, intended specifically for the cluster @lucbelliveau set up for now. Mostly a port of the docker-compose file right now, will need to be adjusted.


gcconnex-kube.yaml
```diff
@@ -0,0 +1,53 @@
...
+          value: "/var/www/html"
+        - name: "DBHOST"
+          value: "gcconnex-db"
+        - name: "WWWROOT"
+          value: "http://localhost:8080/"
+---
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: gcconnexdb-deployment
```
lucbelliveau: Should have a different name than the deployment above.

gcconnex-kube.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: gcconnex-deployment
+  namespace: gctools-test
+spec:
+  replicas: 2 # tells deployment to run 2 pods matching the template
```
lucbelliveau: How are sessions handled in GCconnex?  Are they persisted in the database?  Or are they php's default file based sessions?  If they are in the database all good; otherwise having multiple replcas will cause problems with authentication...  also if there are any other stateful requests  that need to work with temporary files, or anything else that is specific to a single host, then more than 1 replica will blow up.

gcconnex-kube.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - containerPort: 80
+        env:
+        - name: "DOCKER"
+          value: "1"
+        - name: "HOME"
+          value: "/var/www/html"
+        - name: "DBHOST"
+          value: "gcconnex-db"
+        - name: "WWWROOT"
+          value: "http://localhost:8080/"
```
lucbelliveau: Should be the service URL

gcconnex-kube.yaml
```diff
@@ -0,0 +1,53 @@
...
+        image: localhost:5000/gctools-outilsgc/gcconnex:latest
+        ports:
+        - containerPort: 80
+        env:
+        - name: "DOCKER"
+          value: "1"
+        - name: "HOME"
+          value: "/var/www/html"
+        - name: "DBHOST"
+          value: "gcconnex-db"
```
lucbelliveau: The database deployment will need a headless server declaration as well - the name to use here will be `servicename.gctools-test`

gcconnex-kube.yaml
Phanoix: It's different but should probably be more visibly so.

gcconnex-kube.yaml
Phanoix: Should be database from what I remember... I'll double check that.

gcconnex-kube.yaml
lucbelliveau: Oh, no its fine.. just didn't look closely enough

PR #1657, opened 2017-12-15
Rewrote dockerfile to base of alpine

lucbelliveau:
Layers should be optimized for quick upload to remote servers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,60 @@
-FROM ubuntu:14.04
-MAINTAINER Luc Belliveau <luc.belliveau@nrc-cnrc.gc.ca>
-
-# Install dependencies
-ARG DEBIAN_FRONTEND=noninteractive
-RUN apt-get update
-RUN apt-get install -y git apache2 php5 libapache2-mod-php5 php5-mysql php5-gd php5-curl curl
-RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
-
-# Enable Apache mod_rewrite
-RUN a2enmod rewrite
-
-# Modify Apache config for Elgg
-RUN echo '<Directory /var/www/html>\nDirectoryIndex index.php\nOptions FollowSymLinks MultiViews\nAllowOverride All\nOrder allow,deny\nallow from all\n</Directory>\n' | sed '/^<VirtualHost \*:80>/r /dev/stdin' /etc/apache2/sites-available/000-default.conf > /etc/apache2/sites-available/tmp
-RUN mv /etc/apache2/sites-available/tmp /etc/apache2/sites-available/000-default.conf
-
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,33 +1,60 @@
...
+COPY ./install/config/htaccess.dist /var/www/html/.htaccess
+COPY --from=0 /app/vendor/ /var/www/html/vendor/
+COPY . /var/www/html
+
+WORKDIR /var/www/html
+EXPOSE 80
+
 # Start Apache in foreground mode
-CMD chown www-data /var/www/html/data && chown www-data /var/www/html/engine && rm -f /var/run/apache2/apache2.pid && /usr/sbin/apache2ctl -D FOREGROUND
+CMD rm -f /run/apache2/httpd.pid && /usr/sbin/apachectl -D FOREGROUND
```
Phanoix: Looks like you have to add apachectl as a separate package: http://pkgs.alpinelinux.org/contents?file=apachectl

Dockerfile
```diff
@@ -1,33 +1,60 @@
...
+    apache2-utils \
+    php5-apache2 \
+    php5-curl \
+    php5-dom \
+    php5-gd \
+    php5-json \
+    php5-mysql \
+    php5-xml \
+  && mkdir -p /var/www/html/vendor \
+  && mkdir -p /var/www/html/data \
```
Phanoix: I've been meaning to change it in the Dockerfile so might as well do it now, the data dir should really be outside of a directory served by apache. just /data would work.

Dockerfile
lucbelliveau: Yes, it's part of apache2-utils which is installed...  did you get an error?

Phanoix: Wasn't there for me... I'll test again tomorrow.

Dockerfile
```diff
@@ -1,33 +1,59 @@
...
-RUN sed -i '/ErrorLog ${APACHE_LOG_DIR}\/error.log/c\ErrorLog \/dev\/stderr' /etc/apache2/sites-available/000-default.conf
-
-# Uncomment to also see access log (rebuild with docker-compose up --build)
-# RUN sed -i '/CustomLog ${APACHE_LOG_DIR}\/access.log combined/c\CustomLog \/dev\/stdout combined' /etc/apache2/sites-available/000-default.conf
-
-COPY . /var/www/html
-WORKDIR /var/www/html
-
+# First stage, install composer and its dependencies and fetch vendor files
+FROM alpine:latest
```
nschonni: Might be good to pin to a particular version. I mention this because if there are native dependencies I've seen them break from the 3.4 -> 3.6 NodeJs image migration

lucbelliveau: absolutely, I'll make that change.

PR #1693, opened 2018-01-25
Friend_Request Pagination

bmrgould:
Index for Friend_Request uses _list_ over _get_ and was changed to handle that.
Sent and Received pages were completely redone based on the user page so it can handle list.


sdkibb:
"All checks have passed
2 successful checks"? Well then...

Issue #1696, opened 2018-01-29
Closed Groupsstill visible to non-members should still be searchable


klecuyer:
## User Story for the issue in the format of "As a user of type AAA, I need XXX so that I can YYY."
As a user searching for groups, if a group is closed, but visible to non-members, it should appear in searches. Remove noindex tag from closed groups that are not 'invisible'.

Solution: remove following meta tag from group pages:
<meta name="robots" content="noindex, follow">

## Details on issue or enhancement

This is what the group settings appear as in the scenario described above:

![image](https://user-images.githubusercontent.com/29754185/35522115-85b57eb4-04e9-11e8-9e3a-c8e73d40d8db.png)

## For the development team
- [x] Issue user story documented
- [x] UX input received
- [x] Design completed
- [x] Design validated by business team / UX
- [ ] Code review completed by peer
- [ ] Issue closing comment references any duplicate or connected issues or pull requests
- [ ] Issue closed

klecuyer:
@pandurx

pandurx:
see referenced pull request for more information

pandurx:
@klecuyer @4rsood  as discussed, we are looking into this issue. the robot meta tag for noindex has been removed, the gsa should be able to pick up the group profile page

klecuyer:
@pandurx @4rsood thank you. Please keep me posted so I can inform the user.

pandurx:
@klecuyer
@4rsood  has restarted the crawl for all the groups, with the new patch that was implemented on Monday. Confirmed that the group RCMP Policy Community now shows up in the results.

PR #1701, opened 2018-01-31
closed group not being indexed

pandurx:
Issue #1696


pandurx:
*Issue user story documentation*
- group profile had a meta tag noindex that prevented the gsa to crawl their profile
- fix: checks the right field (group visibility) to determine to crawl the group profile

*UX input, Design, Design validation*
- Not available

PR #1703, opened 2018-01-31
uploading zipped file causes fatal error

pandurx:
not directly related to issue #1703


Troy-Lawson:
all this does is remove a call to error log? does that fix the problem?

pandurx:
yes, that resolves the issue

PR #1719, opened 2018-02-13
Add user detail to discussion group api

Steph4104:
Add more data to group discussion 

Also, a test to fix the json code


Steph4104:
@Phanoix , I change the place of the code. Right now, it create the userDetail array, and give the name and description of the discussion without the json code. 

This need to be push on prod of GCcollab so I can test if its work.

markwooff:
@Steph4104 I've placed your `group.php` file on GCcollab prod, test away!

PR #1729, opened 2018-02-22
Check if a first post exists

markwooff:
Fix for https://github.com/gctools-outilsgc/gccollab/issues/276

### DIFF
--- a/mod/gc_communities/views/default/js/stream_community_wire.js
+++ b/mod/gc_communities/views/default/js/stream_community_wire.js
@@ -70,20 +70,22 @@ function load_new_community_wire_items() {
 
 	// Get the guid from the post id
 	var firstPostOnPage = $('.community-wire-holder ul li').first().attr('id');
-	var postID = firstPostOnPage.split("-");
-	postID = postID.slice(2);
+    if (firstPostOnPage) {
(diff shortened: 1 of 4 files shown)


markwooff:
@ethanWallace just added some checks into your streaming content JS files. Also changed the selector since it was causing issues with Wire Post re-shares (see https://github.com/gctools-outilsgc/gccollab/issues/276).

ethanWallace:
@markwooff I tested this on my local and the wire and the department activity pages are working correctly. On the community pages the wire widget in the sidebar is working correctly but the main newsfeed seems to constantly think there is a new post. And finally the News Feed page doesn't seem to work anymore. I checked to see what was wrong and it seems like it cant find any new posts.

markwooff:
Thanks for the comments @ethanWallace . The selector I had used for the newsfeed wasn't always valid so I've replaced it with one that should work regardless of the content type being displayed.

I'm not sure why the community newsfeed is constantly thinking there is a new post, all I've added was an if statement to ensure a post exists from the selector. Plus we don't have this feature on GCcollab to test with. Any ideas?

ethanWallace:
@markwooff I will take a look to see why the community newsfeed is acting strange.

Issue #1649, opened 2017-12-07
Group Admin Options not working


KevinChagnon:
Some of the Group Admin Options are not working;

Limit Owner Block: Does not work - no change on Group home page
Hide the Join Group/Request Membership button:  Does not work - no change on Group home page
Hide Sidebar items:  Does not work - no change on Group home page
Hide Search in Groups:  There is no Search in group

![image](https://user-images.githubusercontent.com/23527358/33726686-bb4a0498-db43-11e7-973d-35503b721b96.png)

Default Sorting Options: Does not work - Selection is saved but folders not sorted according to selection - its displays folders with the default settings

![image](https://user-images.githubusercontent.com/23527358/33726805-12a5c236-db44-11e7-8843-23b09324cc5f.png)

![image](https://user-images.githubusercontent.com/23527358/33726821-2606cf82-db44-11e7-9c50-0918b5ae77f1.png)

KevinChagnon:
https://github.com/Phanoix/gcconnex-help/issues/43

Steph4104:
@KevinChagnon  The default folder sorting option is for file, not folder. Do we want keep it that way and change the language?

KevinChagnon:
As the function is made to sort files and not folders, we will change the language to match the function.

The text should be changed to:

"Default **file** sorting options" for English

"Option de tri de **fichiers** par défaut" for French

Both for the header and title of drop down.

PR #1732, opened 2018-02-23
Fix group admin option

Steph4104:
Change the language for sorting folder to file. Only file is updating, not folder.

Remove search, allow user to join and limit owner block from option.

Fix hide sidebar.

close issue #1649 
close issue Phanoix/gcconnex-help#43

### DIFF
--- a/mod/gc_group_layout/lib/groups.php
+++ b/mod/gc_group_layout/lib/groups.php
@@ -356,7 +356,7 @@ function groups_handle_profile_page($guid) {
 	$content = elgg_view('groups/profile/layout', array('entity' => $group));
 	$sidebar_value = $group->getPrivateSetting('group_tools:cleanup:menu');
 
-	if($sidebar_value == 'yes'){
+	if($sidebar_value == 'no'){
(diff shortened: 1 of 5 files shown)


mod/gc_group_layout/lib/groups.php
```diff
@@ -354,8 +354,12 @@ function groups_handle_profile_page($guid) {
 	groups_register_profile_buttons($group);
 
 	$content = elgg_view('groups/profile/layout', array('entity' => $group));
-	$sidebar = elgg_view('groups/sidebar/sidebar', array('entity' => $group));
-    
+	$sidebar_value = $group->getPrivateSetting('group_tools:cleanup:menu');
+
+	if($sidebar_value == 'yes'){
```
piet0024: This will need to test for 'no'. The form says "hide sidebar items", when I select 'Yes' it displays the items, and when I select 'No' it hides them.

Steph4104: Oups, you are right, I made the change!

PR #1737, opened 2018-02-27
Api for discussion and share

Steph4104:
Update the API to be able to get discussion in group withou json tag and share the content on the wire.

### DIFF
--- a/mod/gc_mobile_api/models/group.php
+++ b/mod/gc_mobile_api/models/group.php
@@ -554,7 +554,17 @@ function get_group_discussions($user, $guid, $limit, $offset, $lang)
 		'order_by' => 'e.last_action desc'
 	));
 
-	return json_decode($discussions);
+	$discussions = json_decode($discussions);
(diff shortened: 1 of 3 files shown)


mod/gc_mobile_api/models/group.php
```diff
@@ -554,7 +554,17 @@ function get_group_discussions($user, $guid, $limit, $offset, $lang)
...
 	));
 
-	return json_decode($discussions);
+	$discussions = json_decode($discussions);
+	foreach ($discussions as $discussion) {
+		
+		$discussion->userDetails = get_user_block($discussion->owner_guid, $lang);
+		$discussion->title = gc_explode_translation($discussion->title, $lang);
+		$discussion->description = gc_explode_translation($discussion->description, $lang);
+		error_log(print_r($discussion->userDetails,true));
```
markwooff: Left an error_log call in.

Steph4104: Oups, I removed

Issue #1699, opened 2018-01-31
500 Error when inviting to group, specific to Career Connexions


klecuyer:
## User Story for the issue in the format of "As a user of type AAA, I need XXX so that I can YYY."
The Career Connexions profile (logged in as) clicking 'Invite users to the group' in the Career Connexions group brings a 500 error. This seems to be specific to this profile/group.

![image](https://user-images.githubusercontent.com/29754185/35628002-9a8e6578-0668-11e8-8f81-bde60d5b1a0e.png)

![image](https://user-images.githubusercontent.com/29754185/35627982-8ee5ba32-0668-11e8-9ee1-00a800f5d752.png)

## Details on issue or enhancement

Link to profile: https://gcconnex.gc.ca/profile/Careerconnexionscarriere
Link to group: https://gcconnex.gc.ca/groups/profile/4174873/career-connexions-carriere

## For the development team
- [ ] Issue user story documented
- [ ] UX input received
- [ ] Design completed
- [ ] Design validated by business team / UX
- [ ] Code review completed by peer
- [ ] Issue closing comment references any duplicate or connected issues or pull requests
- [ ] Issue closed

klecuyer:
Linked to: https://github.com/Phanoix/gcconnex-help/issues/55

smellems:
The `Careerconnexionscarriere` is a generic user account and has 5671 colleagues.  I was able to replicate the bug with all groups owned by that user account. The error on the server is `Out of memory`.

I suspect that it's caused by the high number of colleagues, since that's what's gets listed first when you go to `invite users to the group`.  Tested with a user that has 3000 colleagues and it works.

smellems:
If a user has more that 1000 colleagues, we will display a warning message saying we can't display their colleagues.

@LemieuxGen @KevinChagnon what should the text say?

en: You have too many colleagues to display here!
fr: Vous avez trop de collègues pour les afficher ici!

LemieuxGen:
@smellems, what about this?

EN: Looks like you're too popular! Your colleagues list can't be loaded because you have over 1,000 colleagues. 
FR: Il semble que vous êtes trop populaire! Votre liste de collègues ne peut pas être chargée car vous avez plus de 1 000 collègues.

klecuyer:
@smellems unfortunately it still gives a 500 error
![image](https://user-images.githubusercontent.com/29754185/36980478-634115c0-2058-11e8-9a6c-3a8200d5e5e4.png)

[...]

PR #1739, opened 2018-02-27
Limit colleagues to 1000 when inviting users to group

smellems:
Patch for #1699

### DIFF
--- a/mod/wet4/languages/en.php
+++ b/mod/wet4/languages/en.php
@@ -147,7 +147,7 @@ return array(
 
     'group:accept:invitation' => 'Accept<span class="wb-inv">group invite to "%s"</span>',
     'group:decline:invitation' => 'Decline<span class="wb-inv">group invite to "%s"</span>',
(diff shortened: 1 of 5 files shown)


mod/wet4/languages/en.php
```diff
@@ -147,7 +147,7 @@
 
     'group:accept:invitation' => 'Accept<span class="wb-inv">group invite to "%s"</span>',
     'group:decline:invitation' => 'Decline<span class="wb-inv">group invite to "%s"</span>',
-
+    'groups:toomanyfriends' => 'Looks like you\'re too popular! Your colleagues list can\'t be loaded because you have over 1,000 colleagues.',
```
piet0024: I feel it would be valuable to let the user know they can still use the "Find member" option to invite users. @LemieuxGen

ethanWallace:
@smellens Any chance we will get the changes in today for the push this weekend?

Phanoix:
can this be merged in now?

PR #1742, opened 2018-03-01
Group Mobile API - Enabled Features, Private Desc, Login Location

bmrgould:
For each api function the login check was moved to before getting the entity, deals with errors.

Boolean Enabled variables returned with get_group. They repersent the different group features that can be used, such as group activity, group events, group blogs, etc. Allows mobile app to hide anything not enabled.

Group description based on access.

Tested on collab production with mobile app. 

Different parts were committed/pulled into collab repo separately if you want to read them individually.

### DIFF
--- a/mod/gc_mobile_api/models/group.php
+++ b/mod/gc_mobile_api/models/group.php
@@ -265,17 +265,46 @@ function get_group($user, $guid, $lang)
 	$group->liked = count($liked) > 0;
 
 	$groupObj = get_entity($group->guid);
+	$group->public = $groupObj->isPublicMembership();
 	$group->member = $groupObj->isMember($user_entity);
+	if (!$group->public && !$group->member){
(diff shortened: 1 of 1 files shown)


mod/gc_mobile_api/models/group.php
```diff
@@ -265,17 +265,46 @@ function get_group($user, $guid, $lang)
...
+	$group->public = $groupObj->isPublicMembership();
 	$group->member = $groupObj->isMember($user_entity);
+	if (!$group->public && !$group->member){
+		$group->access = false;
+	} else {
+		$group->access = true;
+	}
+	//Group 'Tools' that are enabled or not
+	//Returning info hide anything not activitated
+	$group->enabled;
```
markwooff: Should initialize this variable to something before setting its sub-objects.

PR #1743, opened 2018-03-01
Merge account feature

ethanWallace:
Gives the ability to site admins to merge one user account into another through the admin panel. When the two accounts are selected you have the option to transfer written content(blogs, discussions etc), colleagues, group membership, profile (education, work experience and skills only), deactivate the old account or delete the old account.

As discussed files and images will not be transferred and are left for the user to transfer themselves.


mod/merge_users/actions/users/merge.php
```diff
@@ -163,6 +200,15 @@
   unset($old_user->gc_skills);
 }
 
+//Test for deactivate
+$deactivate = get_input('deactivate');
+
+if($deactivate){
+  $old_user->gcdeactivate = true;
+  $old_user->gcdeactivatereason = "";
```
piet0024: This can take a string so we can put "merge" or something here so we can keep track.

PR #1747, opened 2018-03-06
add more specification to state

Steph4104:


mod/cp_notifications/start.php
```diff
@@ -710,9 +710,12 @@ function cp_create_annotation_notification($event, $type, $object) {
 	if (strcmp($object_subtype, 'likes') != 0) {
 
 		$content = get_entity($object->entity_guid);
-
+		$get_error_info='if object subtype likes 0';
```
pandurx: i am sort of confused about this part, the condition checks for a string match for likes. 
did you put this here to see where the draft could potentially send out notification?

PR #1750, opened 2018-03-06
URL Language parameter

pandurx:


pandurx:
this has been pushed to production

LemieuxGen:
@pandurx since this change, the language toggle doesn't work anymore on GCconnex content where the URL shows the language parameter. Can you check this out to see how it can be fixed as soon as possible?

Issue #1709, opened 2018-02-05
Fix spelling of "english" in bilingual content toggle


KevinChagnon:
When selection (top of page) is French; The spelling of "english" in the bilingual content toggle for Pages is incorrect. it should be "**E**nglish" ;

![image](https://user-images.githubusercontent.com/23527358/35813447-9f33d8e0-0a61-11e8-9762-0785d33504c9.png)

## For the development team
- [ ] Issue user story documented
- [ ] UX input received
- [ ] Design completed
- [ ] Design validated by business team / UX
- [ ] Code review completed by peer
- [ ] Issue closing comment references any duplicate or connected issues or pull requests
- [ ] Issue closed

KevinChagnon:
https://github.com/Phanoix/gcconnex-help/issues/59

rleir:
Hi Kevin, Gen,
Just a guess, but I bet you dollars-to-donuts that this is the culprit. PR soon.
cheers -- Rick
./mod/wet4/languages/en.php:'box:indicator:en' => 'This content is also available in ',
./mod/wet4/languages/fr.php:    'box:indicator:en' => 'This content is also available in ',

./mod/wet4/languages/en.php:'indicator:click:en' => 'english.',
./mod/wet4/languages/fr.php:    'indicator:click:en' => 'english.',

./mod/wet4/languages/en.php:'indicator:click:fr' => 'français',
./mod/wet4/languages/fr.php:  'indicator:click:fr' => 'français',

Used in this way in many wet views:
<?php echo elgg_echo('box:indicator:en') ?><span class="fake-link" id="fake-link-1"><?php echo elgg_echo('indicator:click:en') ?></span></span>

markwooff:
Fixed with https://github.com/gctools-outilsgc/gcconnex/pull/1753. Thanks @rleir

PR #1753, opened 2018-03-09
capitalize language names for #1709

rleir:
Hi all,
See the comment below. HTH!
Rick


rleir:
Chrome inspect shows me this 
<span class="fake-link" id="fake-link-1">english.</span>
which matches the markup in the wet views.
This is however still untested!

PR #1782, opened 2018-03-22
Opportunity group and level search

### LINKED ISSUE #1305: Add ability to Find Members by group+level in Advanced Find.
In Find Members -> Advanced Find, add search filter field based on group and level

Phanoix:
Group and level search are separate allowing users to search by only group by a range of levels, or a combination thereof.

close #1305


LemieuxGen:
@Phanoix Thanks for working on this. I have a few required changes:
- When searching by classification, it makes sense for users to search by classification only, but not to search by level only. We should have a field for "Classification". Once the user selected that, they are shown value fields: 1) Group; 2) All levels, =, >=, <=; and 3) Level. This can be similar to how the "duration" field works (see image below). For the 2nd value, the user needs to be able to select an option that will look at all the levels. I suggest "All levels" or leaving it blank, which would mean that the 3rd value for Level would not be populated. 
op duration search ex

In French, the field should be "Classification" and the values should be: 1) Groupe; 2) Tous les niveaux, =, >=, <=; and 3) Niveau.

For the "Level" value, we should add the levels: 00 and 20 (also needs to be added to the options when creating an opportunity). 
FYI- There are also a few groups that have odd levels, like U1-U6. I will verify with OCHRO if we need to adjust the group and levels across the OP (when creating opportunities and searching for opportunities). This could be an other Github issue. 

Note: This is based on how the search by classification works on the [GC Jobs website](https://emploisfp-psjobs.cfp-psc.gc.ca/psrs-srfp/applicant/page2440?fromMenu=true&toggleLanguage=en).

Phanoix:
Got it working as a single item, just needs a bit of cleanup. Will let you know when it's up on preprod.

mod/missions/languages/en.php
```diff
@@ -754,7 +754,7 @@
         "missions:average_number_of_applicants" => "Average number of applicants",
 
     "missions:error:department_restricted"  =>  "This opportunity is only available within: %s.<br /> Your department is currently set to: %s.<br /> If this is not correct please update your profile.",
-    "missions:groupandlevel"=>"Group/Classification",
+    "missions:groupandlevel"=>"Group/level",
```
nschonni: Wouldn't it be `Group/Classification and level`?
EX: "Group or Classification and Level", rather than "Group or Level"

Phanoix: Yeah, I asked @LemieuxGen about that (and the French version). It's more of a placeholder / flagging it to be changed since the old version doesn't even mention level.

PR #1807, opened 2018-04-06
fixed album metadata

pandurx:


mod/wet4/views/default/page/elements/head.php
```diff
@@ -254,20 +254,23 @@
...
-		$my_page_entity = $some_post;
-		$page_entity_type = $my_page_entity->getSubtype();
-	} else {
-		$some_post = get_entity($segments[sizeof($segments) - 2]);
-		if ($some_post instanceof ElggEntity) {
-			$my_page_entity = $some_post;
-			$page_entity_type = $my_page_entity->getSubtype();
-		}
-	}
+  $segments = (strpos($_SERVER['REQUEST_URI'], '?') !== false)
```
pandurx: this is the only change that was made

PR #1808, opened 2018-04-06
Post.Blog Mobile API

bmrgould:
API for mobile app to post/save blogs.

Takes a container guid for posting to blogs, otherwise defaults to user.

Validates and/or defaults inputs. Returns errors for missing required fields.

Creates blog object, and sets variables. 

Saves. Errors on failure.

Creates river.

Returns url to blog post on success.

May update again when working on editing blogs/drafts. Working on production gccollab.


mod/gc_mobile_api/models/blog.php
```diff
@@ -244,3 +264,154 @@ function get_blogposts_by_owner($user, $limit, $offset, $lang, $target)
...
+				 } else {
+					 return ("blog/edit/$blog->guid");
+				 }
+
+		 } else {
+			 return elgg_echo('blog:error:cannot_save');
+		 }
+	 }
+
+	 return "end, unknown result";
```
markwooff: Should probably remove this before pushing to production, other than that - looks good!

PR #1809, opened 2018-04-09
Mobile API - Group Blogs + Blog Refactor

bmrgould:
Moved grabbing a group's blog function to the blog.php file and named it based off it's method of grabbing blogs based off the container_guid. Rather than a group specific function in group.php - Allows future use if applicable by not pinning it to be too specific.

Included missing functionality since the group blogs has not been used while other blogs functions have added info/variables.

Created a foreach_blogs function to handle the repeated aspects of the different list of blogs. Reduces repeated code. Makes maintaining and changing easier as it will be only in one location.


mod/gc_mobile_api/models/blog.php
```diff
@@ -69,6 +85,39 @@
  false
 );
 
+function foreach_blogs($blogs, $user_entity, $lang)
```
markwooff: If this will be re-used in other models, can we move to the generic `inc/functions.php` file? Thanks.

PR #1818, opened 2018-04-13
MobileAPI - Edit Blogs

bmrgould:
**Get_Blog_Edit**  
Grabs a a blog post based on user id and blog post id, with only info needed for edit form

**save.blog replace post.blog**
Replaced post.blog with save.blog, which is the same but with a bit of extra handling to allow Save and Edit inside one function.
Adds conditional for grabbing existing blog or creating new blog.
Extra variables relating to holding revision information.
Takes extra input: blog_guid for editing that entity.

### DIFF
--- a/mod/gc_mobile_api/models/blog.php
+++ b/mod/gc_mobile_api/models/blog.php
@@ -404,7 +404,7 @@ function get_blog_edit($user, $guid, $lang)
  return $blog_post;
 }
 
-function save_blog($user, $title, $excerpt, $body, $container_guid, $comments, $access, $status, $lang)
+function save_blog($user, $title, $excerpt, $body, $container_guid, $blog_guid, $comments, $access, $status, $lang)
 {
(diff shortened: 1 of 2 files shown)


mod/gc_mobile_api/models/blog.php
```diff
@@ -344,7 +359,52 @@ function get_blogposts_by_container($user, $guid, $limit, $offset, $lang)
...
+ }
+
+ if (is_callable(array($group, 'getURL'))) {
+	 $blog_post->groupURL = $group->getURL();
+ }
+
+ return $blog_post;
+}
+
+function save_blog($user, $title, $excerpt, $body, $container_guid, $comments, $access, $status, $lang)
```
markwooff: Missing the new `blog_guid` parameter that you just added.

bmrgould: Explains why my current testing is creating a new blog, ha.

Issue #1825, opened 2018-04-23
Blog Excerpt replaced with blog content in edit mode, does not update in view mode


KevinChagnon:
When I edit a blog post, the contents of the Excerpt portion always defaults to the entire blog contents, when in edit mode rather than a excerpt value that may have been previously issued; but does not show the same in view mode. When the Excerpt is updated, these changes are not seen in view mode and reverts to show the entire content of the blog body

KevinChagnon:
https://github.com/Phanoix/gcconnex-help/issues/63

PR #1829, opened 2018-04-24
fix excerpt in blog

Steph4104:
get value of french version and use better var in the form

fix issue #1825 
fix issue Phanoix/gcconnex-help#63

### DIFF
--- a/mod/wet4/views/default/forms/blog/save.php
+++ b/mod/wet4/views/default/forms/blog/save.php
@@ -29,6 +29,8 @@ $json_title = json_decode($vars['title']);
 $json_desc = json_decode($vars['description']);
 $json_exc = json_decode($vars['excerpt']);
 
+error_log($vars['excerpt']);
+
 if ( $json_title ){
   $vars['title2'] = $json_title->fr;
   $vars['title'] = $json_title->en;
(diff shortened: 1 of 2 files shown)


mod/wet4/views/default/forms/blog/save.php
```diff
@@ -29,6 +29,8 @@
 $json_desc = json_decode($vars['description']);
 $json_exc = json_decode($vars['excerpt']);
 
+error_log($vars['excerpt']);
```
pandurx: This should probably be removed

mod/wet4/views/default/forms/blog/save.php
```diff
@@ -40,10 +42,12 @@
 }
 
 if ( $json_exc ){
-  $excerpt2 = $json_desc->fr;
-  $excerpt = $json_desc->en;
+  $vars['excerpt2'] = $json_exc->fr;
+  $vars['excerpt'] = $json_exc->en;
 }
 
+error_log('fr '.$excerpt2." en ".$excerpt);
```
pandurx: This as well

mod/wet4/views/default/forms/blog/save.php
Steph4104: fix

mod/wet4/views/default/forms/blog/save.php
Steph4104: fix

PR #1835, opened 2018-04-26
change to splash page so site uses single language cookie

Troy-Lawson:


pandurx:
I'm testing this on my local on IE11 because Chrome does something special with cookie management. The cookie doesn't seem to be retaining the new changes (I console.log the string and commented out `parent.location.href=frenchlink` so it doesn't clear my log):

`document.cookie = name + "=" +escape( value ) + ";path=/" + ";expires=" + expires_date.toGMTString() + ";domain=$domain";`

I keep getting:

`Cookie: Elgg=62mlgc2co74l1ho8oall7b9h40`

Troy-Lawson:
@pandurx tagging so you see reply

pandurx:
The logic in the code is perfectly fine, but the browser on my machine might behave differently compared to Preprod/Prod

Troy-Lawson:
@pandurx i dont know if it has been pushed to pre-prod if that is where you are testing it

PR #1838, opened 2018-04-26
Forum post (comments and replies) ACL issue

pandurx:


pandurx:
There shouldn't that many replies or comments in the Forum Topics, but I've added a condition so that it doesn't always save the `access_id` whenever the user loads the topic

PR #1840, opened 2018-04-27
Simple saml lang fixes

Troy-Lawson:
fixes to show language toggle to saml related pages. changes made to allow passing a language code in from SP sources and used to create a language cookie on gcconnex


Troy-Lawson:
language changes are all admin panel side

PR #1859, opened 2018-05-16
cookie for language toggle

pandurx:
the domain for the cookie would not work with local instances (site url would be localhost, or local network ip address with gcconnex as the directory), it will display javascript error in the developer's tool (require.js for translations)

@Troy-Lawson what else was not working with the language param in the url or is that already patched?

### DIFF
--- a/mod/gc_lang_url_handler/start.php
+++ b/mod/gc_lang_url_handler/start.php
@@ -25,6 +25,9 @@ function global_url_handler($hook, $type, $returnvalue, $params) {
 	$site_url = preg_replace("(^https?://)", "", $site_url);
 	$site_url = explode("/", $site_url);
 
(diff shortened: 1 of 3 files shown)


Troy-Lawson:
well this pull request doesnt break anything that is not already broken. The language parameter still seems to be able to break the language on the page of the profile. By toggling the language and editing the url lang parameter you can have the profile page display in french but when editing sections of the profile the access drop down and save buttons appear in Engish

pandurx:
@Troy-Lawson 
https://github.com/gctools-outilsgc/gcconnex/pull/1859/commits/58636457317900afbb8bc800239a237cae50ead5
this commit should now fix the language parameter: changes the cookie then refreshes the page

pandurx:
@Steph4104 the splash page should be fixed on this branch

PR #1867, opened 2018-05-30
API for Solr crawling (Alternative)

pandurx:


mod/solr_api/start.php
```diff
@@ -0,0 +1,302 @@
...
+	foreach ($deleted_records as $deleted_record) {
+		error_log("the id: " . $deleted_record->id);
+	}
+
+
+	return false;
+}
+
+
+// api calls with sample set of limit 15
```
Phanoix: Where's the limit 15? most seem to have a limit of 10.

mod/solr_api/start.php
```diff
@@ -0,0 +1,302 @@
...
+	}
+	return $arr;
+}
+
+
+function get_user_list($offset) {
+
+	$site_url = elgg_get_site_url();
+	$query = "	SELECT e.guid, ue.name, ue.username, ue.email, e.type, e.time_created, e.enabled
+				FROM elggusers_entity ue
```
Phanoix: Might not want to hard-code the prefix

mod/solr_api/start.php
pandurx: oh, oops. i forgot to update the comments, i had planned to default to a limit of 15 but realized the results could be too much so i changed it to 10

mod/solr_api/start.php
```diff
@@ -0,0 +1,302 @@
...
+	));
+
+	foreach ($groups as $group) {
+
+		if (is_Json($group->name)) {
+			$name_array = json_decode($group->name, true);
+			$name_array['en'] = str_replace('"', '\"', $name_array['en']);
+			$name_array['fr'] = str_replace('"', '\"', $name_array['fr']);
+		} else {
+			$name_array['en'] = $group->name_array;
```
Phanoix: $group->name_array? Also, gc_explode_translation( $string, $lang ) does most of this already, you can just apply the replace you use to the output of that if needed.

pandurx: this was questionable because i was thinking that the general public can use this plugin for their solr instance (or for whatever reason they want with all the entities in elgg) that function call is either defined in wet4 or gcconnex_theme? and those plugins are tailored for our requirements

Phanoix: Yeah, it's in wet4. The although the en/fr content json string is also part of our modifications.

mod/solr_api/start.php
```diff
@@ -0,0 +1,302 @@
...
+			}
+
+		} else {
+			$title_array['en'] = $entity->title;
+			$title_array['fr'] = $entity->title;
+		}
+
+		if (is_Json($entity->description)) {
+			$description_array = json_decode($entity->description, true);
+			if (!isset($description_array['en']) || !isset($description_array['en'])) {
```
Phanoix: one of them should be fr, or you can just let gc_explode_translation() handle the logic if you want to use it.

Phanoix:
Unless I missed something this looks like it should be fine.

PR #1869, opened 2018-06-11
Api add event

Steph4104:
Api to create an event


Steph4104:
@bmrgould Yes, I can add the new functionality to have event with group and everything around but I think it will be better if I wait until everything is update to version 2.  Instead of adding functionality under version 1.

PR #1878, opened 2018-07-03
Mobile Events API: Offsets, Target, Now To Future

bmrgould:
Events API, changes impact: get_events, get_events_by_owner, and get_events_by_colleagues

As part of checking the `from` variable passed in, will check it compared to `now`. If no from, or it is in the past, will default to using now as the from. Therefore it will never start from the past, only future events.
Direction changed to ASCending rather than descending, to progress to future events, not past events.

Offsets added as variables to be passed in, and to be used in params.

events_by_owner: Changed method from filter/realtionship to just using  a passed in target guid as the container guid. Allows multiple uses, allowing for group and user's as the target, instead of just the user logged in. Has a check to verify target is a user or group.

### DIFF
--- a/mod/gc_mobile_api/models/event.php
+++ b/mod/gc_mobile_api/models/event.php
@@ -56,6 +56,7 @@ elgg_ws_expose_function(
 		"from" => array('type' => 'string', 'required' => false, 'default' => ""),
 		"to" => array('type' => 'string', 'required' => false, 'default' => ""),
 		"limit" => array('type' => 'int', 'required' => false, 'default' => 10),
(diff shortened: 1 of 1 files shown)


bmrgould:
@Steph4104 DESC will start from the 'largest' date. If someone creates an event for the year 2020 it would be the first one shown, and scrolling through them would go from that future date towards the past. This is different than how the website shows it, https://gccollab.ca/event_calendar/list/

The website, and how I think makes more sense, is showing the 'upcoming events' starting from now, and scrolling into the future. That requires ASC.

PR #1900, opened 2018-08-10
move help articles file into the data directory

Phanoix:
To make this work for with the containerized / multiple instance deployments.
fixes gctools-outilsgc/gccollab#367


Phanoix:
Will need to do something about the hard-coded paths in the actions.

PR #1909, opened 2018-09-06
Distributed notification digest sending

Phanoix:
<s>For now it's the idea and some of the code there, the original function might still retain some resemblance to what it is now after this is done.</s>

Ready for some testing now with multiple nodes, something resembling real data.

fixes gctools-outilsgc/gccollab#370


Phanoix:
Thanks, I think all that's left is to test it with several pods and using real / dev data then.

Phanoix:
Worked as expected with real data with multiple pods, but just in case of exact timing issues, adding a delay between initializing the task queue and starting on dequeuing would reduce the chances of duplication I think.

PR #1911, opened 2018-09-10
Allow solr crawler to crawl without logging in the usual way

Phanoix:
The sitemap will take if from here, should work the same a crawling GCconnex with this.
I'll set up https://dev.gccollab.ca to test this.

fixes gctools-outilsgc/gccollab#360


Phanoix:
Tested with @4rsood, looks like this gets the job done.

PR #1916, opened 2018-09-19
Adding meta tags for various pages (collab)

piet0024:
- Added offset url pattern to the no index array
- Added page owner profile url for content / entities that have an owner
- Added entity type 

Closes https://github.com/gctools-outilsgc/gccollab/issues/365

### DIFF
--- a/mod/wet4_collab/views/default/page/elements/head.php
+++ b/mod/wet4_collab/views/default/page/elements/head.php
@@ -164,7 +164,7 @@ wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licenc
 
           $page_entity_type = "";
           if ($my_page_entity instanceof ElggEntity)
-          $page_entity_type = $my_page_entity->getSubtype();
+            $page_entity_type = $my_page_entity->getSubtype();
 
           if(elgg_instanceof($my_page_entity, 'group')){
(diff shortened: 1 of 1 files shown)


mod/wet4_collab/views/default/page/elements/head.php
```diff
@@ -162,12 +162,18 @@
           echo  date ("Y-m-d", elgg_get_excerpt($my_page_entity->time_updated));
           */
 
+          $page_entity_type = "";
+          if ($my_page_entity instanceof ElggEntity)
+          $page_entity_type = $my_page_entity->getSubtype();
```
Phanoix: indent?

mod/wet4_collab/views/default/page/elements/head.php
```diff
@@ -185,23 +185,22 @@
...
-
-          // condition for forums
-          if ($page_entity_type == 'thewire') {
-            $page_entity_type = 'wire';
-          }
-
-          if ($page_entity_type == 'groupforumtopic') {
-            $page_entity_type = 'discussions';
+          switch($page_entity_type) {
+            case 'page_top':
```
nschonni: Since this is the same with next one, you can "fall through".
```
case 'page_top':
case 'page':
  $page_entity_type = 'pages';
  break;
```

PR #1919, opened 2018-09-21
Settings ideas json error

Steph4104:
JSON error in settings of ideas

Fix https://github.com/gctools-outilsgc/gccollab/issues/376

### DIFF
--- a/mod/toggle_language/views/default/toggle_language/toggle_lang.php
+++ b/mod/toggle_language/views/default/toggle_language/toggle_lang.php
@@ -14,6 +14,11 @@
 		}
 
 		$cookie_name = "lang";
+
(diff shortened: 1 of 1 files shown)


mod/toggle_language/views/default/toggle_language/toggle_lang.php
```diff
@@ -14,6 +14,11 @@
 		}
 
 		$cookie_name = "lang";
+
+		if (isset($SESSION['language']) ){
```
Phanoix: do you mean !isset(), or does the bug only pop up when language is not English?

Steph4104: sorry, wrong branch

PR #1931, opened 2018-10-23
redirect error fix for search crawler sitemap

### LINKED ISSUE #1927: Bug in the GCconnex site map mod set up for the Solr crawler
There may be a bug in the GCconnex site map mod that was set up for the Solr crawler.  Some of the crawled URLs have missing information in the Solr index and may have to be because 
the GCconnex application is not rendering the web contents for all the URLs if the request comes from the "solr-crawler" agent.  

This has resulted in some of the groups' and users' contents not being indexed and therefore, are not found in GCtools search results.  As an example a URL with the pattern https://gcconnex.gc.ca/groups/profile/39784216/e-div-prime-purge-epurge?language=en 
is not rendered to the Solr index whereas a URL pattern like this example https://gcconnex.gc.ca/groups/profile/42175342?language=en is rendered properly.

Phanoix:
fixes #1927

### DIFF
--- a/mod/gc_elgg_sitemap/start.php
+++ b/mod/gc_elgg_sitemap/start.php
@@ -205,7 +205,7 @@ function strip_content_hyperlinks_handler($hook, $type, $return, $params) {
 function redirect_group_url($hook, $type, $url, $params) {
 	// from: 192.168.xx.xx/gcconnex/groups/profile/1111/group-name
 	// to: 192.168.xx.xx/gcconnex/groups/profile/1111/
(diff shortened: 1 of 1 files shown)


Phanoix:
@4rsood this seems to fix the problem with content pages not loading for the crawler, you can test it at https://dev.gccollab.ca

4rsood:
@Phanoix difficult to test at https://dev.gccollab.ca since I don't have the crawling problem on gccollab.  Is it possible to apply the change on gcconnex?

PR #1943, opened 2018-11-26
Cometchat cleanup

Troy-Lawson:
Removed comet chat js and css calls and cleaned up some other unused code


nschonni:
Few other hits https://github.com/gctools-outilsgc/gcconnex/search?q=cometchat&type=Code

Phanoix:
Also the relevant issue(s) should be referenced.

PR #1953, opened 2018-12-17
Add a collab dev docker environment install script

Phanoix:
install script can now either install a collab or connex instance based on env.


Phanoix:
Unrelated issue I found testing this: hard-coded table prefix at https://github.com/gctools-outilsgc/gcconnex/blob/master/mod/cp_notifications/start.php#L302

PR #1957, opened 2019-01-08
Performance enhancements to recommendation provider API

lucbelliveau:
These are performance and minor feature upgrades to the recommendation provider API.

The API can now run under 64M, whereas before it required gigs to run due to how elgg performs queries internally; as well as how the objects were stored in memory before being converted to JSON.  Conversion is now done inline by the generator.

### DIFF
--- a/mod/missions/api/v0/api.php
+++ b/mod/missions/api/v0/api.php
@@ -1,5 +1,4 @@
 <?php
-
 /**
 * NRC Recommendation Platform API Library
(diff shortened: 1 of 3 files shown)


lucbelliveau:
@Phanoix give me a few hours before merging - I am writing some SQL modifications to speed things up...

lucbelliveau:
Ok done.  Took a bit longer than I thought!

PR #1961, opened 2019-01-18
Update geds user search to "begins with"

Troy-Lawson:
Update geds user search to "begins with" as opposed to using "contains". Hopefully leads to users finding their GEDS profile more often


Troy-Lawson:
This doesn't seem to fix the problem yet

Troy-Lawson:
@Phanoix this is now good to review

PR #2009, opened 2019-05-01
limit community wire queries and remove pagination

Phanoix:
It's one of the slowest queries on the platform if not limited.


Phanoix:
Will be removing the filtered wire instead, closing in favour of that.

PR #2022, opened 2019-07-02
Remove tabs in member page

Steph4104:
For issue # https://zube.io/tbs-sct/gctools/c/4968

I put in comment code for removing Recent and online tabs. 
This is for test only. 

When we are sure we want to remove it, I will create a last commit to erase those commit.


Steph4104:
@piet0024 , you are right, on GCconnex, the c_member_byDepartment overwrite members plugin. Should be fix now.

PR #2025, opened 2019-07-04
OpenID new scopes

ethanWallace:
Added new address and phone scopes to pleio mod to gain additional information to build user profile.  If user has a profile picture uploaded on Directory, it will save the url to metadata and use that url to load the profile instead.


Phanoix:
Directory isn't a thing anymore so closing this

PR #2026, opened 2019-07-04
Remove all site activity feed

Steph4104:
Related to https://zube.io/tbs-sct/gctools/c/4965

To remove all site activity feed we need to remove / comment some code outside of the mod repo.

We can only remove the link to the feed from the main menu, but user can still access the page if the type the url.

If we want to remove the entire page, we can only remove the body of the page so the user don't end up on a white page. The user still can access the menu to navigate somewhere else.

### DIFF
--- a/engine/lib/river.php
+++ b/engine/lib/river.php
@@ -867,9 +867,6 @@ QUERY;
  * @access private
  */
 function _elgg_river_init() {
-	elgg_register_page_handler('activity', '_elgg_river_page_handler');
-	//$item = new \ElggMenuItem('activity', elgg_echo('activity'), 'activity');
-	//elgg_register_menu_item('site', $item);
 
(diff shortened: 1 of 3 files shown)


nschonni:
It's better to delete code than leave commented out code sitting around. It can always be brought back from source control

engine/lib/river.php
```diff
@@ -868,8 +868,8 @@ function _elgg_river_enable($event, $type, $entity) {
  */
 function _elgg_river_init() {
 	elgg_register_page_handler('activity', '_elgg_river_page_handler');
```
Phanoix: If you want to remove the page entirely you would want to get rid of the page handler as well, otherwise it should only be the menu item. Also agree on the deleting instead of commenting out since this is all in git.

Phanoix:
done in SGDT branch

PR #2028, opened 2019-07-04
Remove wire sidebar from community pages

piet0024:
Part of zube card 4966. Removed the sidebar that renders the wire.


Phanoix:
There's also the widget (widgets/filtered_wire_index/) and the ajax file (ajax/community_wire.php) which should just be deleted.

Issue #2036, opened 2019-07-24
 Preview Blog Post function broken


Menrika:
## As a GCcollab user
I want a preview function
so that I can preview my post before submitting.
 
Issue:
Currently, using Firefox when I select the preview button it actually publishes the article.
 
Hack:
The current hack is to publish the article as a draft to preview it before posting.
 
 Ticket : https://gccollab.gctools-outilsgc.ca/a/tickets/9590

## For the development team
- [ ] Issue user story documented
- [ ] UX input received
- [ ] Design completed
- [ ] Design validated by business team / UX
- [ ] Code review completed by peer
- [ ] Issue closing comment references any duplicate or connected issues or pull requests
- [ ] Issue closed

PhilAndrecheck:
@Steph4104 qui fait la revue ?

Steph4104:
@PhilAndrecheck , c'était Ethan, je ferme l'issue car la pull request a été merger

PR #2042, opened 2019-08-06
Fixing blog preview

Steph4104:
If preview is click, a cookie is create so modal of gc tags won't show up and blog will be create as a
draft.

Fixing issue https://github.com/gctools-outilsgc/gcconnex/issues/2036


ethanWallace:
I think the way you solved this problem works well. The only thing I think needs to be looked at again is the ability to access the tags modal when publishing the blog. With this change a user will never see that modal.

PR #2062, opened 2019-09-13
View profile

Steph4104:
Clean the view file by removing all comment code and extra file.

Changing a little bit the look to match the new one.

Related to: https://zube.io/tbs-sct/gctools/c/5149


mod/b_extended_profile/start.php
```diff
@@ -153,8 +153,9 @@ function init_ajax_block($title, $section, $user) {
             break;
     }
 
-    echo '<div class="panel panel-custom">';
-        echo '<div class="panel-heading profile-heading clearfix"><h2 class="profile-info-head panel-title pull-left clearfix">' . $title . '</h2>'; // create the profile section title
+    echo '<div class="panel">';
+    echo'<div class="panel-body">';
```
piet0024: can you write a style for this div to padding-bottom: 0. The padding between the title of the card and the body content is too large.

Steph4104: Create profile-style class in gcconnex-profile.css to get the 0 padding at the title.

PR #2131, opened 2019-11-04
Update event calendar

Steph4104:
Update all information from event calendar to the description (and merge room and venue together) to math the new layout of event calendar.

### DIFF
--- a/mod/wet4/views/default/plugins/wet4/settings.php
+++ b/mod/wet4/views/default/plugins/wet4/settings.php
@@ -50,7 +50,22 @@ if ( $page_mode == 'bilingual_upgrade' ) {
 	));
 
 	access_show_hidden_entities($access_status);
-}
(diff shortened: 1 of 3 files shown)


mod/wet4/views/default/plugins/wet4/settings.php
```diff
@@ -50,7 +50,22 @@
...
+}else if($page_mode == 'event_calendar_update'){
+	echo "<br /><h3>upgrade event calendat description</h3>";
+	echo elgg_view("output/url",
+		array(
+			'href' => elgg_get_site_url().'admin/plugin_settings/wet4',
+			'text' => '<< back to settings page',
+		)
+	);
+	echo "<br />";
+	$count = get_data_row("SELECT count(distinct guid) as objects FROM `elgg_entities` where `subtype` = 17");
```
Phanoix: you have a hard-coded table prefix for elgg_entities here

PR #2150, opened 2019-11-18
use custom list_river without pointless count

### LINKED ISSUE #2147: group profile activity module running a count with pagination disabled

Phanoix:
Looks like it's the same thing I was trying to solve with the newsfeed page so it's just reusing the same solution.

closes #2147


Phanoix:
I originally fixed this with a custom elgg_list_group_river, but that still has the pointless count which can take just as long if not longer than the query for the activity, so the solution will be to update that one to only count if pagination is true.

Issue #1982, opened 2019-02-21
XSS in profile page --- Security Vulnerability on GCcollab


Menrika:
## Ticket : https://gccollab.gctools-outilsgc.ca/a/tickets/7349 

" I don't know if you are interested about security vulnerabilities on your GCcollab website but I have found some of them.
 
You can look at the pictures for an example and to see the area where they can be injected.
 
Description
 
Editing Basic Profile page, you can :
 
- Insert a javascript protocol instead of a website http link
    
    javascript:alert('xss')
 
- Insert a crafted link as a social media account
    yo'" onmouseover=alert(`xss`);//
    
    Really bad one :    
    yo'" onmouseover=alert(`xss`) style=display:block;position:fixed;width:9999px;height:9999px;top:0;left:0;z-index:99999;
 
 
I did not look at the code, but it might be elgg not making good checks. "
 
![image](https://user-images.githubusercontent.com/45214337/53180910-899ada80-35c4-11e9-8394-ee836a050d48.png)

## For the development team
- [ ] Issue user story documented
- [ ] UX input received
- [ ] Design completed
- [ ] Design validated by business team / UX
- [ ] Code review completed by peer
- [ ] Issue closing comment references any duplicate or connected issues or pull requests
- [ ] Issue closed

PR #2168, opened 2019-11-25
adding htmlspecialchars() in profile

Steph4104:
Adding htmlspecialchars() when saving content in profile to prevent HTML Injection

Related to: #1982


mod/b_extended_profile/actions/b_extended_profile/edit_profile.php
```diff
@@ -127,10 +127,13 @@
                       $user->set($f, $v);
 
                     } else {
+                        if($f == 'website'){
+                            if ($v && !preg_match("#^((ht|f)tps?:)?//#i", $v)) {
+                                $v = "http://$v";
```
Phanoix: I'd say make it https

PR #2171, opened 2019-11-27
Email analytics

Phanoix:
Email counts in memcached.

### DIFF
--- a/mod/cp_notifications/start.php
+++ b/mod/cp_notifications/start.php
@@ -150,6 +150,17 @@ function minor_save_hook_handler($hook, $type, $value, $params) {
     return true;
 }
 
+function count_emails( $type ){
+	try {
+		$m = new Memcached();
+		$m->setOption(Memcached::OPT_BINARY_PROTOCOL, true);
+		$m->addServer('localhost', 11211);
+
+		$n = $m->increment("email:$type", 1, 1);
+	} catch(Exception $err){
+		error_log("could not increment email counter for $type");
+	}
+}
 
 /**
  *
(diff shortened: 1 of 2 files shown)


mod/cp_notifications/start.php
```diff
@@ -150,6 +150,17 @@ function minor_save_hook_handler($hook, $type, $value, $params) {
     return true;
 }
 
+function count_email_types( $type ){
+	try {
+		$m = new Memcached();
+		$m->setOption(Memcached::OPT_BINARY_PROTOCOL, true);
+		$m->addServer('localhost', 11211);
```
nschonni: should these be config variables?

mod/phpmailer/start.php
```diff
@@ -250,11 +250,17 @@ function phpmailer_send($to, $to_name, $subject, $body, array $bcc = NULL, $html
 		$errType = "custom";
 		phpmailer_logging($errMess, $errStack, 'PHPMailer', $errType);
 	}
-
+	count_emails();
 	return $return;
 }
 
+function count_emails(){
```
nschonni: This is vendor'd in now, but is there a way to do this without modifying the 3rd party code through hooks instead?

Phanoix: Pretty sure only the core elgg mail functionality has a hook but I don't think it would be all that useful with the whole notification / email system being handled by the cp_notifications mod (and missions mod does its own thing I think, will need to add counts there too).

Phanoix:
This was mostly for a quick investigation if I remember right, haven't needed it in the years since so closing this.

PR #2190, opened 2019-12-04
Tab order and focus state fixes

### LINKED ISSUE #2180: Highlight CKEditor section
Adding highlight on focus of the CKEditor section (Reply, comment description).

image

### LINKED ISSUE #2181: Tabs order in replies section is wrong
When using the tab to navigate the replies list, the order start with "profile, menu entities and username" instead of "Profile, username and entity menu"

image

ethanWallace:
Fixed tab order on discussion replies. Closes #2181 
Added focus state to ckeditor. Closes #2180

### DIFF
--- a/mod/wet4/views/default/wet4_theme/custom_css.php
+++ b/mod/wet4/views/default/wet4_theme/custom_css.php
@@ -2066,5 +2066,5 @@ figcaption{
 .cke_focus .cke_contents {
     border-color: #269abc;
     outline: 0;
-    box-shadow: 0 0 3px rgba(0,0,0,.2), 0 0 3px rgba(38,154,188,.7) !important;
+    box-shadow: 0px 0px 8px 2px rgba(102,175,233,.6), 0 0 8px rgba(102,175,233,.6) !important;
 }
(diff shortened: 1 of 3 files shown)


mod/wet4/views/default/wet4_theme/custom_css.php
```diff
@@ -2061,4 +2061,10 @@
 
 .wb-slc {
     height: 0;
+}
+
+.cke_focus .cke_contents {
+    border-color: #269abc;
+    outline: 0;
+    box-shadow: 0 0 3px rgba(0,0,0,.2), 0 0 3px rgba(38,154,188,.7) !important;
```
Steph4104: Everithing is good and work but my suggestion will be to change  the highlight of the focus to something like that: box-shadow: 0px 0px 8px 2px rgba(102,175,233,.6), 0 0 8px rgba(102,175,233,.6).  It has the same blue has other focus and is more "perceptible".

ethanWallace: Sounds good to me

Issue #2192, opened 2019-12-05
GCcollan event UI issue


marc-donofrio:
When attempting to add an event, some users are confused that you need to click on **View all** before the event calendar appears:

## Details on issue or enhancement
![image](https://user-images.githubusercontent.com/42844328/70265236-f94f1000-1767-11ea-9bfe-a3619fd49a36.png)

## For the development team
- [ ] Issue user story documented
- [ ] UX input received
- [ ] Design completed
- [ ] Design validated by business team / UX
- [ ] Code review completed by peer
- [ ] Issue closing comment references any duplicate or connected issues or pull requests
- [ ] Issue closed

PR #2195, opened 2019-12-09
Add "create event" on profile

Steph4104:
Adding the "add event" button in event tabs on profile.

On GCconnex and GCcollab

Fix issue #2192

### DIFF
--- a/mod/b_extended_profile/views/default/profile/tab-content.php
+++ b/mod/b_extended_profile/views/default/profile/tab-content.php
@@ -4,7 +4,7 @@
 */
 //generate content tabs
 elgg_push_context('profile');
-$fields = array('File', 'Blog', 'page_top', 'Bookmarks', 'Poll', 'Thewire', 'Album', 'task_top', 'question');
+$fields = array('File', 'Blog', 'page_top', 'Bookmarks', 'Poll', 'Thewire', 'Album', 'task_top', 'question', 'events');
 $user_display_name = elgg_get_page_owner_entity()->name;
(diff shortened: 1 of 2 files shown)


mod/b_extended_profile_collab/views/default/profile/tab-content.php
```diff
@@ -122,7 +132,15 @@
             }
         }
 
-        if(!$content){
+        if($field == "event_calendar"){    
+            $events = event_calendar_get_personal_events_for_user(elgg_get_page_owner_guid(), 5);
+            if(!$events){
+                echo '<div class="mrgn-lft-sm mrgn-bttm-md">' . elgg_echo('event_calendar:no_events_found11') . '</div>';
```
ethanWallace: There is an extra 11 at the end of the no_events_found in both files. Everything else looks good though.

mod/b_extended_profile/views/default/profile/tab-content.php
```diff
@@ -119,8 +130,15 @@
 
             }
         }
-
-        if(!$content){
+        if($field == "event_calendar"){    
+            $events = event_calendar_get_personal_events_for_user(elgg_get_page_owner_guid(), 5);
+            if(!$events){
+                echo '<div class="mrgn-lft-sm mrgn-bttm-md">' . elgg_echo('event_calendar:no_events_found11') . '</div>';
```
ethanWallace: 11 also seems to be on this no_events_found as well.

Issue #1016, opened 2017-04-27
Don't use the `<i>` element for icons


nschonni:
Although the Font Awesome docs use this, the tag does have semantic meaning https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i
Bootstrap (and WET) switched to `<span>` for those http://getbootstrap.com/components/#glyphicons
This may come more from the Elgg core, I just noticed this looking at some of the code.

ToferC:
Suggest this be part of best practices and review of plugins for content refresh.

nschonni:
Nice 👍

PR #2254, opened 2020-03-17
Changing i tags to span for all icon

Steph4104:
Updating all i tags for span when it use for icon.

Related to : #1016


mod/multi_file_upload/themes/fa/theme.js
```diff
@@ -29,10 +29,10 @@
         },
         previewZoomButtonIcons: {
             prev: '<i class="fa fa-caret-left fa-lg"></i>',
```
piet0024: Do these icons need to be changed to spans?

PR #2258, opened 2020-03-18
Name input label on group create form

piet0024:
While going through the create content forms I noticed the name autocomplete input for groups was missing a label.


Steph4104:
Do we need to update the ID of the input to match the for attribute of the label?

piet0024:
When I was doing my testing, I found the autocomplete input appends `_autocomplete` to the ID var, so the label and input ID will match when the input is rendered on the page.

PR #2263, opened 2020-03-31
minor css fix to allow lists in discussion reply

Troy-Lawson:
small css tweak so that unordered lists appear in discussion reply's properly. Lists will still not appear properly in river calls. that should be made a separate issue


Troy-Lawson:
need to take a second look

PR #2304, opened 2020-08-31
Extend edit profile with paas mutation

ethanWallace:
Extends the edit profile action with the graphql mutation to paas. If graphql mutation fails, the edit profile action will be canceled.

Also merged master branch into this branch so sorry for the mess.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,10 @@ RUN apk --no-cache add \
   php5-zlib \
   php5-curl \
   curl
-RUN mkdir /app && mkdir /app/pleio && mkdir /app/paas_integration && curl -sS https://getcomposer.org/installer | php5 -- --install-dir=/usr/local/bin --filename=composer
+RUN mkdir /app && \
+    mkdir /app/pleio && \
+    mkdir /app/paas_integration && \
+    curl -sS https://getcomposer.org/installer | php5 -- --install-dir=/usr/local/bin --filename=composer
 RUN ln -s /usr/bin/php5 /usr/bin/php
 WORKDIR /app
 COPY composer.json composer.json /app/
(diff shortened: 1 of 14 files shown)


Dockerfile
```diff
@@ -13,11 +13,12 @@ RUN apk --no-cache add \
   php5-zlib \
   php5-curl \
   curl
-RUN mkdir /app && mkdir /app/pleio && curl -sS https://getcomposer.org/installer | php5 -- --install-dir=/usr/local/bin --filename=composer
+RUN mkdir /app && mkdir /app/pleio && mkdir /app/paas_integration && curl -sS https://getcomposer.org/installer | php5 -- --install-dir=/usr/local/bin --filename=composer
```
nschonni, line 16:
```suggestion
RUN mkdir /app && \
        mkdir /app/pleio && \
        mkdir /app/paas_integration && \
        curl -sS https://getcomposer.org/installer | php5 -- --install-dir=/usr/local/bin --filename=composer
```

mod/paas_integration/start.php
```diff
@@ -0,0 +1,104 @@
...
+
+    $variables = array(
+        'gcID' => $gcID,
+        'data' => array(
+            'name' => $profile_fields['name'],
+            'titleEn' => copy_value_to_object_if_defined('job', $profile_fields['job']),
+            'titleFr' => copy_value_to_object_if_defined('jobfr', $profile_fields['jobfr']),
+            'mobilePhone' => copy_value_to_object_if_defined('mobile', $profile_fields['mobile']),
+            'officePhone' => copy_value_to_object_if_defined('phone', $profile_fields['phone']),
+            'address' => array(
```
piet0024, line 76: If the address field is left empty, the profile service throws back  something along the lines of `400 Bad Request ... got invalid value`. However the other fields can be empty and everything fires properly.

piet0024 (changes requested):

piet0024 (approved):

PR #2313, opened 2020-09-16
Maintain approval workflow with PaaS

ethanWallace:
When a user modifies their job titles, the profile page will now query PaaS to inform the user that they change is waiting on approval.

This notice can be found in the edit profile modal.

### DIFF
--- a/mod/paas_integration/start.php
+++ b/mod/paas_integration/start.php
@@ -128,3 +128,54 @@ function edit_profile_paas_mutation($hook, $entity_type, $returnvalue, $params) 
 
     return null;
   }
+
+  function waiting_on_approval($guid){
+
+    $dbprefix = elgg_get_config("dbprefix");
+    $service_url = elgg_get_plugin_setting("graphql_client", "paas_integration");
+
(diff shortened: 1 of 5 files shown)


mod/paas_integration/start.php
```diff
@@ -127,4 +127,55 @@ function address_validation($address) {
...
     return null;
+  }
+
+  function waiting_on_approval($guid){
+
+    $dbprefix = elgg_get_config("dbprefix");
+    $service_url = elgg_get_plugin_setting("graphql_client", "paas_integration");
+
+    $result = get_data_row("SELECT pleio_guid FROM {$dbprefix}users_entity WHERE guid = $guid");
+    if ($result->pleio_guid) {
```
Phanoix, line 138: While I can't think of any way this would be false without data corruption or something like that (connection / db errors should error out before reaching here) maybe this would work better as something like

```php
    if ($result->pleio_guid) {
       // error handling
       return
    }
     $gcID = $result->pleio_guid;
```
to cover that case earlier than when it's sent to paas as an empty / uninitialized variable, assuming I didn't miss something.

Phanoix (approved):

PR #2445, opened 2022-08-17
notification settings link on the notifications page.

### LINKED ISSUE #2425: Add a link to notification settings on the notifications page
On the notifications page in GCcollab add a hyperlink to the notifications settings page.

Link to a user's settings: https://gccollab.ca/messages/notifications/Firstname.Lastname(username)

Link to user's notification settings: https://gccollab.ca/settings/notifications/Firstname.Lastname(username)

doug0102:
fixes #2425

### DIFF
--- a/mod/cp_notifications/languages/en.php
+++ b/mod/cp_notifications/languages/en.php
@@ -20,7 +20,7 @@ $english = array(
 	'minor_save:description' => "Posting new content will send out notifications to those who are subscribed. As the group owner or operator, you can decide not to send out notifications about the new content you post in the group. To do so, select the option “Do not send a notification” below.",
 	'minor_save:checkbox_label' => " Do not send a notification",
 
-	'cp_notifications:name' => "Notification Email",
+	'cp_notifications:name' => "Notification settings",
 	'cp_notification:save:success' => "Settings have been saved successfully",
 	'cp_notification:save:failed' => "Settings did not save successfully",
 
--- a/mod/wet4/views/default/page/layouts/elements/header.php
+++ b/mod/wet4/views/default/page/layouts/elements/header.php
@@ -49,7 +49,7 @@ if ($title || $buttons) {
             ));
         }
         if($checkPage == 'messages') {
-            $notificationSettings = elgg_echo('messages:notificationsettings');
+            $notificationSettings = elgg_echo('cp_notifications:name');
 
             $notificationsSettingLink = "{$_SERVER['SERVER_NAME']}/settings/notifications/";
             
--- a/mod/wet4/views/default/page/layouts/elements/header.php
+++ b/mod/wet4/views/default/page/layouts/elements/header.php
@@ -56,17 +56,15 @@ if ($title || $buttons) {
 
             $notificationsSettingLink =  elgg_get_site_url() . "settings/notifications/{$username}";
             
-            $notificationSettingsBtn = "<ul class=\"elgg-menu elgg-menu-title list-inline pull-right elgg-menu-title-default\" style=\"padding-left: 5px\"><li class=\"elgg-menu-item-add\"><a href=\"{$notificationsSettingLink}\" class=\"elgg-menu-content btn btn-primary btn-md\">{$notificationSettings}</a></li></ul>";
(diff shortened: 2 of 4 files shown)


mod/wet4/views/default/page/layouts/elements/header.php
```diff
@@ -48,6 +48,13 @@
             'class' => 'list-inline mrgn-rght-sm',
             ));
         }
+        if($checkPage == 'messages') {
+            $notificationSettings = elgg_echo('cp_notifications:name');
+
+            $notificationsSettingLink = "{$_SERVER['SERVER_NAME']}/settings/notifications/";
```
Phanoix, line 54: elgg has a built-in function for the root url, elgg_get_site_url()
also if you wanted to add the username here elgg_get_logged_in_user_entity() will get you that, and that object has a ->username property, as in
```php
$user_object = elgg_get_logged_in_user_entity();
$username = $user_object->username
```

in this case though it seems to work without the username but this will probably be useful later

Phanoix:
Not sure about  notification settings being a button right beside "compose a message", someone with a more designer~y eye than me might be better taking a look at that.

mod/wet4/views/default/page/layouts/elements/header.php
doug0102: Oh thanks for that tip, didn't know about that built in function. I've updating the PR with the changes you suggested. I will reach out to Nick and see what he thinks about where I but the link.

mod/cp_notifications/languages/en.php
```diff
@@ -20,7 +20,7 @@
 	'minor_save:description' => "Posting new content will send out notifications to those who are subscribed. As the group owner or operator, you can decide not to send out notifications about the new content you post in the group. To do so, select the option “Do not send a notification” below.",
 	'minor_save:checkbox_label' => " Do not send a notification",
 
-	'cp_notifications:name' => "Notification Email",
+	'cp_notifications:name' => "Notification settings",
```
piet0024, line 23: I think we may have a language string that is translated for notification settings. https://github.com/gctools-outilsgc/gcconnex/blob/master/mod/wet4/languages/en.php#L742
try `notifications:usersettings`

mod/wet4/views/default/page/layouts/elements/header.php
```diff
@@ -64,6 +74,6 @@
             }
         }
         $format_title = elgg_view_title($vars['title'], array('class' => 'elgg-heading-main mrgn-lft-sm'));
-        echo elgg_format_element('div', ['class' => 'd-flex title-button-combo'], $format_title .'<div class="title-action-button d-flex">' . $buttons2 . $buttons . '</div>');
+        echo elgg_format_element('div', ['class' => 'd-flex title-button-combo'], $format_title .'<div class="title-action-button d-flex">' . $buttons2 . $buttons . $notificationSettingsBtn . '</div>');
```
piet0024, line 77: I think the position of the settings button is good. Maybe we can add a `btn-default` style class it make it look like a secondary button. This way the main action visually still looks like compose.

mod/cp_notifications/languages/en.php
doug0102: @piet0024 Part of me renaming that from email to settings was Ilia found that the title of the notification settings page was incorrect in English. It was using that language string, so I fixed the English version only. There is a ticket created for it #2446

mod/wet4/views/default/page/layouts/elements/header.php
Phanoix: I think that looks a lot better, the only question I have now is: do we want to keep the settings link on all of the messages pages (inbox, notifications, and messages received)? Personally I think it's fine since it makes about as much sense as having compose message on all of them too and I don't think it gets in the way really.

piet0024: Yes I think it's fine as is. I think having the link to the settings available on the page is overall a positive UX fix to direct users to their settings. I'll approve this PR.

piet0024 (approved):

PR #2450, opened 2022-08-23
Changed h2 header for colleague component so it includes the profile …

### LINKED ISSUE #2434: GCcollab Profile - Rename colleague sidebar to [Page owner entity]'s colleagues
**Consistency and standards**
When viewing another user’s profile, their colleague show up as “My colleagues” which is confusing to the user. Change to “[Name]’s Colleagues” so as to be clear which user it relates to.

The code for this should be in the `b_extend_profile_collab` mod.

Image

LecCory:
…owner's name

resolves #2434

### DIFF
--- a/mod/wet4/views/default/page/components/module.php
+++ b/mod/wet4/views/default/page/components/module.php
@@ -143,6 +143,16 @@ if( $type == 'GPmod'){
         $header = elgg_format_element('h2', ['class' => 'panel-title'], $title);
     }
 
+    // CL 23082022 - Coll
+    $owner = elgg_get_page_owner_entity()->username;
+    if ($title === 'Colleagues') {
+        $header = elgg_format_element('h2', ['class' => 'panel-title'], "$owner's $title");
+    }
+    if ($title === 'collègues') {
+        $header = elgg_format_element('h2', ['class' => 'panel-title'], "$title à $owner");
+    }
+    
+
     if ($header !== null) {
         $header = elgg_format_element('header', ['class' => 'panel-heading'], $header);
     }
--- a/mod/wet4/views/default/page/components/module.php
+++ b/mod/wet4/views/default/page/components/module.php
@@ -143,13 +143,14 @@ if( $type == 'GPmod'){
         $header = elgg_format_element('h2', ['class' => 'panel-title'], $title);
     }
 
-    // CL 23082022 - Coll
+    // CL 23082022 - component determines if the h2 title is Colleagues or collegues. Appends the pager owner 
     $owner = elgg_get_page_owner_entity()->username;
     if ($title === 'Colleagues') {
-        $header = elgg_format_element('h2', ['class' => 'panel-title'], "$owner's $title");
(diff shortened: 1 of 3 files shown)


mod/wet4/views/default/page/components/module.php
```diff
@@ -143,6 +143,16 @@
...
     }
 
+    // CL 23082022 - Coll
+    $owner = elgg_get_page_owner_entity()->username;
+    if ($title === 'Colleagues') {
+        $header = elgg_format_element('h2', ['class' => 'panel-title'], "$owner's $title");
+    }
+    if ($title === 'collègues') {
+        $header = elgg_format_element('h2', ['class' => 'panel-title'], "$title à $owner");
+    }
```
doug0102, line 153: I tested this and it seems like when you change the page to French it doesn't pickup the profile name, it just says "Collègues". I think it may be overwritten somewhere else since "Collègues" is capitalized but the FR string you edited is lowercase "collègues". You might also want to switch the second if statement to an else if.

doug0102 (changes requested):

mod/wet4/views/default/page/components/module.php
LecCory: That's strange since it's changing for me in my dev env. I will however, fix the capitalization for both the Title and the page owner. I know the caching is a bit of a pain with Elgg and I've had to clear cache and refresh a couple times before my changes were reflected
image

doug0102: Hmm, I tried flushing the cache and I get the same result. I put a console log in the code and it does pick up my name and title so I have no clue why it's only showing Collègues for the French version. @piet0024 can you take a look? 

image

mod/wet4/views/default/page/components/module.php
```diff
@@ -143,6 +143,17 @@
         $header = elgg_format_element('h2', ['class' => 'panel-title'], $title);
     }
 
+    // CL 23082022 - component determines if the h2 title is Colleagues or collegues. Appends the pager owner 
+    $owner = elgg_get_page_owner_entity()->username;
```
piet0024, line 147: I can see the header display in French in my environment. 
image
I'm not sure why it won't display for Shea. I'll keep looking into this.

For review, I would update `username` here to just `name`. This should display the user's preferred display name set in their profile. Usernames on the platform are system generated so I think displaying `name` here would be best.

piet0024 (changes requested):

mod/wet4/views/default/page/components/module.php
doug0102: I pulled your new changes and can see now that the English version has a capitalized first name meaning my code is up to date. I still have no name for the French version. Not sure what's going on, but since it's just on my side I will resolve this. Might want to keep an eye out to see how it behaves in prod though.

mod/wet4/views/default/page/components/module.php
LecCory: Swapped username to the name prop
image
image

piet0024 (approved):
I reviewed the changes locally, looks good!

Issue #2453, opened 2022-08-26
Update my groups page headings for a11y navigation


piet0024:
During a11y testing the user suggested we update our page heading structure to allow for easier navigation. Look to add a visually hidden h2 for "my groups" just before the groups list. Put the group list links in h3.

![Image](https://user-images.githubusercontent.com/7603237/186508968-80f46a48-c776-4a03-aaba-b9b2952eb22a.PNG)

LecCory:
Group link wrapped in h3, hidden h2 now available when the active link is My Groups. Uses the groups:yours key from the mod/groups/language files

![Image](https://user-images.githubusercontent.com/17506947/188962115-a08ed3a8-5840-4cb8-8f16-4b1fee3e290f.png)

PR #2458, opened 2022-09-07
added h3 tags to activity card summaries and hidden h2 for smoother screen reader nav

### LINKED ISSUE #2452: Wrap activity river item content in h3 element for a11y navigation
During a11y testing the user suggested that we wrap the river activity items in h3 elements. This will allow users with screen readers to navigate quickly through their activity by using these headings.
Visually the style of the posts should remain the same but be wrapped in h3 elements (highlighted in image below).
This should be applied to all instances of the river activity (home page newsfeed index widget, activity widgets, and group activity). Ensure that the beginning of the activity feed starts with an h2 elements letting user's know this is an activity feed. The h2 can read "activity feed" and if the text is not needed on the page the heading can have the style class `wb-inv` to be read only by assistive tech.

Image

LecCory:
resolves #2452 & #2453


doug0102:
Looks good to me

doug0102 (approved):

PR #2466, opened 2022-11-10
fix for smb network file shares + apache

Phanoix:
azure smb file share mounting + ubuntu 18 + apache is a fun combination that by default prepends html headers to files it serves from the network drive, going to be trying to fix it for good here.
Looks like for elgg the problem has nothing to do with apache and everything to do with the combination of sendfile() following ob_end_clean() being broken specifically when reading the files from an smb mounted azure file store regardless of the actual server output buffering settings.


vid:
😵‍💫

PR #2475, opened 2022-12-13
Fixed Members page

### LINKED ISSUE #2474: GCcollab - Members page update
Image

Image

LecCory:
fixes #2474 

Removed the Friends/Colleagues button
Changed the title of the Members menu item to Network/Réseau
Changed the URL for Network/Réseau to: `/friends/<username> `which is the page that the former Friends/Colleagues button navigated to

image


doug0102:
I'm seeing the button removed in the top tight, but the main menu seems different than the screen shot. I attached images of my screen as well as the branch/commit I'm currently on.

image
image

Phanoix:
The nav menu is different from other menus in that it's generated based on active mods as well as being configurable in admin/appearance/menu_items  on the admin side so it's not going to always look the same unless you go there and configure it to be like prod for example. Elgg also by default caches views and language files so you'll need to flush (admin/dashboard) or turn off (admin/settings/advanced) your local dev cache to see changes to those.

mod/wet4/views/default/navigation/menu/elements/item.php
```diff
@@ -49,6 +55,14 @@
 	$roles = "role=\"$item_role\" ";
 }
 
+// CL 20221213 - Get the menu title, if matches Network or Réseau change the url to friend/$username instead of /members
+// CL 20221213 - re: https://github.com/gctools-outilsgc/gcconnex/issues/2474
+if($item->getText() == 'Network' || $item->getText() == "Réseau"){
```
Phanoix, line 60: There's a less hacky way that this can be done in elgg even with how much of a mess the we4 mod is:
First, removing the members menu item, the mod already removes a lot so this fits right in:
`elgg_unregister_menu_item('site', 'members');`
with this bunch  https://github.com/gctools-outilsgc/gcconnex/blob/master/mod/wet4/start.php#L258


and the colleagues / friends menu item already gets registered further down in the same file, so that can just be adjusted to instead add it to the site menu instead over here:
https://github.com/gctools-outilsgc/gcconnex/blob/master/mod/wet4/start.php#L538

Phanoix (changes requested):
This does work but there's a simpler or at least more direct and correct elgg way to manage menu items.

mod/gccollab_theme/views/default/page/elements/user_menu.php
```diff
@@ -99,14 +99,7 @@
...
         </li>
-        <li class="elgg-menu-item-colleagues">
-            <?php
-                // colleagues
-                echo '<a href="/friends/'.$username.'" title="'.elgg_echo("userMenu:colleagues").'" class="elgg-menu-content">
-                    <span class="fa fa-users fa-lg"><span class="wb-invisible">'.elgg_echo("userMenu:colleagues").'</span></span>'.$friend_badge.'
-                </a>';
-            ?>
-        </li>
+        <!-- CL 20221213 - Removed Friends Nav button re: https://github.com/gctools-outilsgc/gcconnex/issues/2474 -->
```
nschonni, line 102: Probably don't need a comment, since the git history covers that with a good commit message

Phanoix (approved):

PR #2510, opened 2023-03-24
First Iteration of GCCollab Main Navigation Live Updates 

AdiMakkar:


Phanoix (approved):
the user menu accessibility will need some work next, can't get to it with keyboard only

PR #2545, opened 2023-05-23
Update site-brand.php

piet0024:
Quick style patch, I will create a ticket for a proper look at the elements.


Phanoix:
it seems to break drop-downs - both in the site menu for the career one, and the user one on the right side. Search drop-down at least looks to not be broken.

PR #2550, opened 2023-05-30
add memcached to app container

Phanoix:
Getting dev and PR test/review instances automatically set up to actually connect to this will require #2409


LecCory (changes requested):
Package is installing, however the process isn't automatically starting. 
Might be worth adding `/etc/init.d/memcached start` somewhere in the Dockerfile to get the process to spin up with the container. 

edit: specified wrong directory

LecCory (approved):

LecCory:
I'll close my open PR that I opened to fix the memcached startup

Issue #2573, opened 2023-07-10
Create Opportunity: Change time commitment input


doug0102:
![image](https://github.com/gctools-outilsgc/gcconnex/assets/2327968/ccf1bcbe-a60e-48ea-9c25-09632a67f834)
![image](https://github.com/gctools-outilsgc/gcconnex/assets/2327968/876961a5-ddaf-48ad-bc51-92a23a96555e)
![image](https://github.com/gctools-outilsgc/gcconnex/assets/2327968/4138eae7-0fdb-4f2a-a70c-d4895df2fc74)
![image](https://github.com/gctools-outilsgc/gcconnex/assets/2327968/50d6ccae-3276-4dbb-b65b-262a52456b03)

PR #2587, opened 2023-07-12
Change time commitment input

AdiMakkar:
PR removed.


AdiMakkar:
For requirements check #2573

PR #2604, opened 2023-09-09
none required removed when no schedule is provided

### LINKED ISSUE #2580: Opportunity View: Remove "None required" from post when no schedule is provided
image
image

AdiMakkar:


mod/missions_collab/views/default/page/elements/print-mission-more.php
```diff
@@ -282,9 +282,11 @@
 			<h5><?php echo elgg_echo('missions:timezone') . ':';?></h5>
 			<span name="mission-time-zone"><?php echo $clean_timezone;?></span>
 		</div>
-		<div class="col-sm-offset-1">
-			<?php echo $unpacked_time; ?>
-		</div>
+        <?php if((!empty($mission->mon_start)) && (!empty($mission->tue_start)) && (!empty($mission->wed_start)) && (!empty($mission->thu_start)) && (!empty($mission->fri_start)) && (!empty($mission->sat_start)) && (!empty($mission->sun_start))) : ?>
```
Phanoix, line 285: it's possible that only some days are filled out, and I'd assume we should be showing the schedule in that case too, so it should be checking for at least one of these not being empty.

Phanoix (dismissed):

PR #2605, opened 2023-09-09
user department advanced search field removed

### LINKED ISSUE #2576: Find Members Advanced Search: Remove "User department" field
image
image
image

AdiMakkar:
closes #2576


mod/missions/views/default/forms/missions/advanced-search-form.php
```diff
@@ -23,7 +23,7 @@
 		elgg_echo('missions:experience'),
 		elgg_echo('missions:skill'),
 		elgg_echo('missions:portfolio'),
-		elgg_echo('missions:user_department'),
```
Phanoix, line 26: I wonder if this is something we want to duplicate / override in the missions_collab mod to not potentially affect connex, it's simple enough so might as well?

AdiMakkar: I am not sure if it's working. Would need a second look from you.

Phanoix (changes requested):
should be in `mod/missions_collab/views/default/forms/missions/advanced-search-form.php`

Phanoix (approved):

PR #2615, opened 2023-09-28
embedded icon file title-alt text updated

### LINKED ISSUE #2581: Opportunity View: Add label to file attachments
image
image

AdiMakkar:
closes #2581


AdiMakkar:
The scope of the above ticket is updated to add an alt-text requirement to the embedded file icon, ensuring accessibility. The ability to add text does exist in the description box and can be used to add text as the user wants.

Phanoix (approved):
This also fixes alt / title text in all file lists!
Note: because of the hacky way the embedding works, it takes the title / alt test from the En/Fr title based on the language of the user who embeds the file.

PR #2686, opened 2024-11-25
adds an option to disable notifications for new non-missions content

### LINKED ISSUE #2677: GCconnex Decommissioning - Prevent system notifications
Leaving any notifications required for careers marketplace and the system to functions properly such as password resets.

Easiest way is probably to disable them in https://github.com/gctools-outilsgc/gcconnex/blob/master/mod/cp_notifications/start.php

Phanoix:
Making this a more formal option, since the way notifications are coded right now can cause serious slowdowns and timeouts, so even outside the gcconnex decommission context this might be useful to have as an option.

Added a placeholder message for the notification settings page since that seems like a reasonable place to mention if this has been activated.

closes  #2677

### DIFF
--- a/mod/cp_notifications/languages/en.php
+++ b/mod/cp_notifications/languages/en.php
@@ -5,6 +5,7 @@ $site_name = $site->name;
 $contact_us = "{$site->getURL()}mod/contactform/?utm_source=notification_digest&utm_medium=email";
 
 $english = array(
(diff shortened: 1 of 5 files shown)


Phanoix:
not worth it right now, would be about time for read only mode by the time we would put it up anyways.

PR #2690, opened 2024-12-04
profile contents completed

### LINKED ISSUE #2678: GCconnex Decom - Read Only Mode Mod for Profile Editing Content Types
Read Only Mode Mod for all profile editing content types. Include profile components: 

- [x] Avatar
- [x] About Me
- [x] Education
- [x] Work Experience 
- [x] Skills

Leave CM Opportunities and Opt Ins.

AdiMakkar:
closes #2678


Phanoix:
the b_extended_profiles directory with the profile edit views we're overriding here is not in /views/default/forms/b_extended_profiles  but only in /views/default/b_extended_profiles

Phanoix:
Also, this seems to do nothing to skills, but since that's part of the careers marketplace overall functionality we wouldn't want to touch that part yet anyways, so best to just not include the skills file at all I'd say.

Phanoix (approved):
Now it seems to work as expected.

PR #2699, opened 2024-12-17
Updated the about us language

AdiMakkar:


Phanoix:
Since we're updating this, might as well set 'expages:gcconnex_demenage' in en.php as well since the way elgg language work with English as the default, it'll default to English if the French doesn't exist but won't do it the other way around.

mod/gcconnex_theme/languages/en.php
```diff
@@ -54,7 +54,7 @@
 	 * Decommission
 	 */
 	'wet:footDecomm' => 'GCconnex is Moving', 
-	'expages:gcconnex_is_moving' => 'GCconnex is Moving to GCXchange!', 
+	'expages:gcconnex_demenage' => 'GCConnex is Moving to GCXchange! | GCConnex passe à GCÉchange!',
```
Phanoix, line 57: expages:gcconnex_is_moving still needs to be there for the English page

Phanoix (changes requested):

Phanoix (approved):

PR #2702, opened 2025-01-12
Read only banner

### LINKED ISSUE #2696: GCconnex Decom - Update the design and comms pieces for the content types (except CM)
Update the design and comms for all banners for all content types (except CM)

AdiMakkar:
closes #2696


Phanoix:
the image should probably be in the read only mod along with everything else, unless it's going to be used a lot in the wet4 mod too.

mod/gcconnex_read_only/start.php
```diff
@@ -17,13 +17,14 @@ function read_only_mode(){
 function decommission_message(){
     echo "<div class='panel panel-default'>";
```
Phanoix, line 18: this can be just a single echo of a multi-line string, in php simply 
```php
echo "...
...
...";
```
should work fine, or alternatively have the function return the string and echo it out wherever it's needed as just `echo decommission_message()`

AdiMakkar: Added a single echo for now. The decommission message is about to be same for the logging in landing page but that will change for CM.

Phanoix (approved):

PR #2710, opened 2025-02-06
Read only mode CM

### LINKED ISSUE #2682: GCconnex Decom - Read Only Mode Mod for the Career Marketplace Types
Read Only Mode Mod for all content types for the career marketplace and types skipped for previous types.

AdiMakkar:
closes #2682


Phanoix (approved):
Works well from what I can tell, see my other comment for something that we may or may not want to add since it might only be visible in a small minority of cases and those buttons won't work anyways in full read only mode.

Phanoix:
Is completely removing the list of applicants how we want to go about it instead of just removing the buttons? Seems to me like it's removing information that some people might look for but if this is what we're going with then this does do it.

mod/gcconnex_read_only_CM/views/default/profile/wrapper.php
```diff
@@ -0,0 +1,227 @@
...
+
+        if ( has_content($user, 'work') ) {
+            init_ajax_block_no_edit_CM(elgg_echo('gcconnex_profile:experience'), 'work-experience', $user);
+            //echo elgg_view('b_extended_profile/work-experience');
+            finit_ajax_block('work-experience');
+        }
+
+        if ( has_content($user, 'gc_skills') && (strstr(strtolower($_SERVER['HTTP_USER_AGENT']), 'gsa-crawler') === false && strstr(strtolower($_SERVER['HTTP_USER_AGENT']), 'solr-crawler') === false) ) {
+            init_ajax_block_no_edit_CM(elgg_echo('gcconnex_profile:gc_skills'), 'skills', $user);
+            //echo elgg_view('b_extended_profile/skills');
```
Phanoix, line 93: not sure if this was left commented out by accident but we probably don't want to have content disappear, only the edit buttons, same for the other sections.

Phanoix (dismissed):

mod/gcconnex_read_only_CM/views/default/profile/wrapper.php
AdiMakkar: changes to the content have been removed, reverted back to only edit buttons - left commented out by accident.

AdiMakkar:
Screenshot 2025-02-19 at 11 29 26 AM

Phanoix (approved):

AdiMakkar:
Changes approved by @buryva, merging things in.

PR #2713, opened 2025-02-13
Export group email list functionality for group admins

Phanoix:
User owners and operators will be able to get a list of user emails, useful for when the email members functionality crashes due to too many members, this should work for groups with up to 50 000 users without optimizations to reduce memory usage.


AdiMakkar (approved):

AdiMakkar:
closes #2716

Steph4104:
Hey good morning, here's the button text with translation:
 
EN: Members Email List
 
FR: Liste de courriels des membres

Menrika (approved):

PR #2737, opened 2026-07-17
remove italics from discussion notifications

doug0102:


Phanoix (approved):
yep, that got rid of the description italicization

